PR #1, opened 2018-01-14
Fix checkstyle error

slachiewicz:
s


hboutemy:
-1 as I feared, this way of writing the package-info.java just breaks the intent = having a package documentation in javadoc output
try to build the javadoc, and you'll see that the "fix" in fact breaks the package javadoc

slachiewicz:
@hboutemy you're right - I'm sorry. 

I try to compile and publish javadocs and in my configuration (Java9 i had problems with maven-enforcer-plugin/extra-enforcer-rules - they failed with EnforceBytecodeVersion exceptions - works fine with latest snapshot versions).

Finally when i upgrade to maven-checkstyle-plugin 3.0.0 (and maven-javadoc-plugin for sure) all works fine for me - without modification to code.

PR #2, opened 2018-07-26
Bug/MSHARED-416

KathrynN:
Fixed MSHARED-416 (tested on Windows and MacOS) where arguments with uneven numbers of ' failed to execute in the shell.

### DIFF
--- a/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
+++ b/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
@@ -19,22 +19,27 @@ package org.apache.maven.shared.utils.cli;
  * under the License.
  */
 
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
+
 import org.apache.maven.shared.utils.Os;
 
-import junit.framework.TestCase;
+import org.junit.Test;
 
 public class CommandLineUtilsTest
-    extends TestCase
 {
 
     /**
      * Tests that case-insensitive environment variables are normalized to upper case.
      */
+    @Test
     public void testGetSystemEnvVarsCaseInsensitive()
     {
         Properties vars = CommandLineUtils.getSystemEnvVars( false );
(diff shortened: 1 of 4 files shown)


khmarbaise:
Lets see what CI says......

khmarbaise:
Would you please follow the contribution guideline and create a single commit with:
```
[MSHARED-416] - Summary from Jira
Optional descriptions
```
I appreciate to merge this into master...

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -19,22 +19,29 @@
...
 import java.util.Properties;
+import java.util.concurrent.TimeUnit;
+
 import org.apache.maven.shared.utils.Os;
 
-import junit.framework.TestCase;
+import org.junit.Assert;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
khmarbaise: Would please so nice to keep explicit import instead of asteric imports...Thanks.

khmarbaise:
CI looks good: https://builds.apache.org/job/maven-box/job/maven-shared-utils/job/MSHARED-416/

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
KathrynN: Done!  Thanks for catching it.

KathrynN:
Updated commit message and rebased on new master as requested - thanks for the fast response!

khmarbaise:
Great Kathryn, I'm just rechecking against CI ....

khmarbaise:
Thanks Kathryn. Merged in master. Thanks for your contribution.

khmarbaise:
The last commit was done with a different email address than the previous ones so I can't see the image in master branch...

KathrynN:
What does that mean and how/can I fix it?

khmarbaise:
Hi Kathryn unfortunately you/I can't fix that anymore cause it is in master ..we can't overwrite master...it is like is...it's not really an issue. I just realized that there was something different.

KathrynN:
Odd...

khmarbaise:
Only the one which is associated with your GitHub account ..one option (usually not a good idea)...better to configure the email address in Git for that particular repository... 
Since Git 2.13.0 there is nice feature which help a lot about that:

https://blog.github.com/2017-05-10-git-2-13-has-been-released/#conditional-configuration
That helped me a lot...cause I have several projects and dozens of repos...without that it would be impossible to handle that...

khmarbaise:
If you have further question just get into the Maven IRC chat ...or on the dev/users list....

jira-importer:
Resolve #223

PR #6, opened 2019-06-04
[MSHARED-822] MessageBuilder should support colored Paths

mthmulders:
Together with Robert Scholte we figured that the idea of hyperlink output is not that easy. Instead, we came with a different idea to color file paths. It is depicted in [the JIRA issue](https://issues.apache.org/jira/browse/MSHARED-822?focusedCommentId=16855592&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16855592).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,9 @@
     <contributor>
       <name>Kathryn Newbould</name>
     </contributor>
+    <contributor>
+      <name>Maarten Mulders</name>
+    </contributor>
   </contributors>
 
   <properties>
--- a/src/main/java/org/apache/maven/shared/utils/logging/AnsiMessageBuilder.java
+++ b/src/main/java/org/apache/maven/shared/utils/logging/AnsiMessageBuilder.java
@@ -19,7 +19,6 @@ package org.apache.maven.shared.utils.logging;
  * under the License.
  */
 
-import org.apache.maven.shared.utils.StringUtils;
 import org.fusesource.jansi.Ansi;
 
 import java.io.File;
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -64,6 +64,7 @@
     <checkstyle.violation.ignore>RedundantThrows,NewlineAtEndOfFile,ParameterNumber,MethodLength,FileLength,ModifierOrder</checkstyle.violation.ignore>
 
     <mavenVersion>3.0</mavenVersion>
+    <javaVersion>7</javaVersion>
```
michael-o: This really be a separate issue.

src/main/java/org/apache/maven/shared/utils/logging/Style.java
```diff
@@ -31,15 +31,18 @@
...
+    DEBUG(       "bold,cyan"   ),
+    INFO(        "bold,blue"   ),
+    WARNING(     "bold,yellow" ),
+    ERROR(       "bold,red"    ),
+    SUCCESS(     "bold,green"  ),
+    FAILURE(     "bold,red"    ),
+    STRONG(      "bold"        ),
+    MOJO(        "green"       ),
+    PROJECT(     "cyan"        ),
+    WORKING_DIR( "black"       ),
```
michael-o: That is going to be hard if your terminal is black.

src/test/java/org/apache/maven/shared/utils/logging/AnsiMessageBuilderTest.java
```diff
@@ -108,4 +113,52 @@ public void should_color_project_and_reset()
...
+        sut.path( absolutePath );
+
+        String expectedPath =  workingDir.toString() + File.separator;
+        assertThat(sut.toString(), containsString("\u001B[30m" + expectedPath + "\u001B[0;34m"));
+    }
+
+    @Test
+    public void should_color_module_dir_from_path()
+    {
+        Path filePath = Paths.get("core","src", "main").toAbsolutePath();
```
michael-o: nitpick. space missing.

src/main/java/org/apache/maven/shared/utils/logging/Style.java
mthmulders: Fair enough, didn't think of that. Not specifying a style here gives you black letters on a light terminal, and white letters on a black terminal. How about that?

pom.xml
mthmulders: Then we should do that first, and it would block this issue. We use the `java.nio.file` in this pull request so we can't continue before that is resolved.

mthmulders:
We used solely system properties first, but it didn't work. Investigated together with @rfscholte and it turns out that some in Maven distributions the `maven.multiModuleProjectDirectory` system property is not present while the  `MAVEN_PROJECTBASEDIR` environment variable is.

src/main/java/org/apache/maven/shared/utils/logging/Style.java
rfscholte: We need to experiment with this to see what looks best. I've been thinking about green / blue / bold+blue.
I think we should stick to 2 colors, making 1 bold as well.

rfscholte:
> Together with Robert Scholte we figured that the idea of hyperlink output is not that easy. Instead, we came with a different idea to color file paths. It is depicted in [the JIRA issue](https://issues.apache.org/jira/browse/MSHARED-822?focusedCommentId=16855592&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16855592).

Yes, hyperlinks won't work, nor would colors. It requires a different MessageBuilder next to AnsiMessageBuilder, something like HtmlMessageBuilder. Title is still valid: paths should have a separate display method.

pom.xml
michael-o: Correct, please do so!

michael-o:
The `maven.multiModuleProjectDirectory` has been there for a long time. Can't we use fallthrough?

pom.xml
mthmulders: See [MSHARED-826](https://issues.apache.org/jira/browse/MSHARED-826). I'll pick that one up soon.

src/main/java/org/apache/maven/shared/utils/logging/Style.java
mthmulders: Sure, we could also restrict ourself to using less colours. I don't have a strong preference, so whatever you think is best suited...

pom.xml
mthmulders: That one is now merged, and I removed the commits in this PR after having rebased the branch on master.

src/main/java/org/apache/maven/shared/utils/logging/AnsiMessageBuilder.java
```diff
@@ -203,6 +207,8 @@ public AnsiMessageBuilder format( String pattern, Object... args )
     @Override
     public String toString()
     {
-        return ansi.toString();
+        String result = ansi.toString();
+        System.out.println( result );
```
michael-o: Why is this nessary? All output is written through SLF4J.

mthmulders: It's a work-in-progress as I explained in the commit message. I want to create a few screenshots in different environments so we can easily see what the outcomes will look like.

mthmulders:
I've attached a few screenshots with different colouring schemes on two types of terminals to the JIRA issue.

mthmulders:
> The `maven.multiModuleProjectDirectory` has been there for a long time. Can't we use fallthrough?

It wasn't always set properly, but I think with Root-reactor aware subfolder builds in Maven 4, it may be more reliable...

mthmulders:
No immediate need for this, but it would be a nice little gem. The current implementation might need some refactoring, moving logic out of the `AnsiMessageBuilder` to keep that one clean.

PR #7, opened 2019-06-05
[MSHARED-826] Require Java 7

mthmulders:
As suggested by @rfscholte in [MSHARED-826](https://issues.apache.org/jira/browse/MSHARED-826): updated to maven-parent:33 and set property `javaVersion` to 7.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,6 +145,13 @@
           </excludes>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <reuseForks>false</reuseForks>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <groupId>org.apache.maven.shared</groupId>
     <artifactId>maven-shared-components</artifactId>
-    <version>32</version>
+    <version>33</version>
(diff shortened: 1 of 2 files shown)


rfscholte:
The OS related tests need to be changed, see https://builds.apache.org/job/maven-box/job/maven-shared-utils/job/MSHARED-826/test_results_analyzer/
With Java 12 you cannot change the System properties anymore that were provided by Java at startup.

Tibor17:
@rfscholte 
The Java 12 has a new API to override system properties at runtime?

rfscholte:
@Tibor17 No, there's just a set of properties that are marked as static properties: they can only be set at startup and cannot be changed afterwards. https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/jdk/internal/util/StaticProperty.java shows a list.
However, it doesn't mention the os properties... might need some extra analysis.

mthmulders:
> The OS related tests need to be changed

I assume this is also the case on master? Since this PR doesn't touch a single test...

Tibor17:
@mthmulders 
One thing is to set higher Java 1.7 and another thing is to use Java 1.7 features in Java code.
You know the features of Java 1.7.
Try to open another PR where you complete this improvement in code; otherwise bytecode version does not have any value for nobody. Thx for your work!

mthmulders:
Thanks for the suggestion. I deliberately didn't want to mix the two (requiring Java 1.7 and using its new features). Java 7 is required for [MSHARED-822](https://issues.apache.org/jira/browse/MSHARED-822) / #6, so we'll soon make use of at least some of the new features.

jira-importer:
Resolve #281

PR #17, opened 2020-03-08
[MSHARED-861] remove unreachable package

elharo:
@michael-o None of the model classes in the reflection package are public. Thus unless someone is doing some really skanky reflection hacks that won't work in java 9+, this is completely unused. My best guess is that someone meant to do something with this, but abandoned it before making it public.


michael-o:
This requires a JIRA issue first.

michael-o:
Did you actually try the tests on Java 11+? Did they fail?

jira-importer:
Resolve #275

PR #19, opened 2020-03-08
remove info about what's not here

elharo:
@michael-o package info only described classes that weren't present


jira-importer:
Resolve #288

PR #21, opened 2020-03-11
[MSHARED-862] system independent file separator

elharo:
@rfscholte see if we can make DirectoryScannerTest pass on Windows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ out/
 /bootstrap
 /dependencies.xml
 .java-version
+.checkstyle
+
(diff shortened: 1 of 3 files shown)


michael-o:
Let me run this too...

elharo:
Interesting. I've never seen that one. Do you also see that one in master?

michael-o:
> 
> 
> Interesting. I've never seen that one. Do you also see that one in master?

Yes, same on master.

elharo:
Build is blue (green) so I'm going to merge.

michael-o:
Ouch, you just have provided 6 junk commits on master. Please avoid that by squashing first. No merge commit since they don't add any benefit for us.

elharo:
Squash commits have a different negative side effect that Robert has asked me to avoid. You should probably bring this up on the dev mailing list, so the community can decide what they want. When a decision is made, the repos can be configured to make the preferred way the only way.

michael-o:
Where did he request this?

elharo:
Slack + email. I don't have a major preference here one way or the other, but I am getting contradictory requests about merging from different committers and PMC members, so let's figure out what everyone is most comfortable with, document that decision, and configure the repos to make it the default or only option.

jira-importer:
Resolve #284

PR #22, opened 2020-03-21
docs: we don't shade commons-io

elharo:
@hboutemy 
Maybe we used to, but not anymore.


hboutemy:
you're right, it was changed for maven-shared-utils 3.0.0 https://issues.apache.org/jira/browse/MSHARED-442

PR #25, opened 2020-04-04
[MSHARED-866] don't throw a java.lang.Error when reading invalid or corrupt properties files

elharo:
@michael-o


jira-importer:
Resolve #291

PR #28, opened 2020-05-03
[MSHARED-884] - Don't always overwrite filtered resources

roxspring:
- Added unit tests for `FileUtils.copyFile()` methods using `FilterWrapper`
- When filtering and _potentially_ overwriting an existing file:
  - Open the existing file as a `RandomAccessFile`
  - Encode each buffer of content
  - Compare encoded buffer with a matching buffer from the existing file
  - If different then overwrite the remainder of the file  
- Added unit tests to demonstrate that lastModified date is only updated when the content changes.
- Rudimentary performance tests with various source sizes show no significant impact:

  Source Size | Avg Millis (Old) | Avg Millis (New)
  -- | -- | --
  100 MB | 6034 | 5940
  1 MB | 54 | 56
  10 KB | 1 | 8

Relates to MSHARED-884

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
@@ -1810,10 +1810,16 @@ public class FileUtils
                                  @Nullable FilterWrapper[] wrappers, boolean overwrite )
         throws IOException
     {
-        if ( wrappers != null && wrappers.length > 0 )
+        if ( wrappers == null || wrappers.length == 0 )
         {
-            
-            if ( encoding == null || encoding.isEmpty() ) 
+            if ( overwrite || to.lastModified() < from.lastModified() )
+            {
+                copyFile( from, to );
+            }
+        }
+        else
+        {
+            if ( encoding == null || encoding.isEmpty() )
             {
                 encoding = Charset.defaultCharset().name();
             }
@@ -1831,13 +1837,6 @@ public class FileUtils
                 }
 
                 IOUtil.copy( wrapped, fileWriter );
-            }
-        }
-        else
-        {
-            if ( to.lastModified() < from.lastModified() || overwrite )
-            {
-                copyFile( from, to );
             }
         }
     }
--- a/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
@@ -40,6 +40,7 @@ import java.io.RandomAccessFile;
 import java.io.Reader;
 import java.io.Writer;
 import java.net.URL;
+import java.nio.Buffer;
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.channels.FileChannel;
@@ -1868,7 +1869,7 @@ public class FileUtils
                         int n;
                         while ( -1 != ( n = wrapped.read( newChars ) ) )
                         {
-                            newChars.flip();
+                            ( ( Buffer ) newChars ).flip();
 
                             coderResult = encoder.encode( newChars, newBytes, n != 0 );
                             if ( coderResult.isError() )
@@ -1876,13 +1877,13 @@ public class FileUtils
                                 coderResult.throwException();
                             }
 
-                            newBytes.flip();
+                            ( ( Buffer ) newBytes ).flip();
 
                             if ( !writing )
                             {
                                 existingRead = existing.read( existingBytes.array(), 0, newBytes.remaining() );
-                                existingBytes.position( existingRead );
-                                existingBytes.flip();
+                                ( ( Buffer ) existingBytes ).position( existingRead );
+                                ( ( Buffer ) existingBytes ).flip();
 
                                 if ( newBytes.compareTo( existingBytes ) != 0 )
                                 {
@@ -1899,9 +1900,9 @@ public class FileUtils
                                 existing.write( newBytes.array(), 0, newBytes.remaining() );
                             }
 
-                            newChars.clear();
-                            newBytes.clear();
-                            existingBytes.clear();
+                            ( ( Buffer ) newChars ).clear();
+                            ( ( Buffer ) newBytes ).clear();
+                            ( ( Buffer ) existingBytes ).clear();
                         }
 
                         if ( existing.length() > existing.getFilePointer() )
--- a/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
@@ -36,11 +36,16 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.io.RandomAccessFile;
 import java.io.Reader;
 import java.io.Writer;
 import java.net.URL;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.charset.Charset;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CoderResult;
 import java.nio.file.Files;
 import java.security.SecureRandom;
 import java.text.DecimalFormat;
@@ -110,7 +115,7 @@ public class FileUtils
     /**
      * The file copy buffer size (30 MB)
      */
-    private static final long FILE_COPY_BUFFER_SIZE = ONE_MB * 30;
+    private static final int FILE_COPY_BUFFER_SIZE = ONE_MB * 30;
 
     /**
      * The vm line separator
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java
```diff
@@ -1,29 +1,49 @@
...
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeFalse;
+import static org.junit.Assume.assumeThat;
+import static org.junit.Assume.assumeTrue;
```
roxspring, line 46: Looks like Intellij moved some imports when I wasn't looking. Happy to fix if it's an issue.

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -1826,34 +1831,101 @@ public static void copyFile( @Nonnull File from, @Nonnull File to, @Nullable Str
                                  @Nullable FilterWrapper[] wrappers, boolean overwrite )
         throws IOException
     {
-        if ( wrappers != null && wrappers.length > 0 )
+        if ( wrappers == null || wrappers.length <= 0 )
```
roxspring, line 1834: Flipped the condition around to get the simple case out of the way and ease the cognitive load when reading the else clause. Maybe I should split this out into a separate commit and make that final check `==` since negative lengths shouldn't be possible!

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -109,7 +114,7 @@ protected FileUtils()
     /**
      * The file copy buffer size (30 MB)
      */
-    private static final long FILE_COPY_BUFFER_SIZE = ONE_MB * 30;
+    private static final int FILE_COPY_BUFFER_SIZE = ONE_MB * 30;
```
roxspring, line 117: Don't think there was any need for this to be a `long` other than the only previous use case, where it's now automatically cast to a long anyway.

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -1826,34 +1831,101 @@ public static void copyFile( @Nonnull File from, @Nonnull File to, @Nullable Str
...
+                    }
+                }
+                else
+                {
+                    CharsetEncoder encoder = Charset.forName( encoding ).newEncoder();
+
+                    int totalBufferSize = FILE_COPY_BUFFER_SIZE;
+
+                    int charBufferSize = ( int ) Math.floor( totalBufferSize / ( 2 + 2 * encoder.maxBytesPerChar() ) );
+                    int byteBufferSize = ( int ) Math.ceil( charBufferSize * encoder.maxBytesPerChar() );
```
roxspring, line 1872: This is attempting to consume the same amount of buffer memory in total, but split between a `char` buffer and a pair of `byte` buffers - each big enough to contain the encoded `char` buffer.

src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java
```diff
@@ -438,9 +448,300 @@ public void copyFile1()
             testFile1.lastModified() == destination.lastModified());*/
     }
 
+    private static long MILLIS_PER_DAY = 24 * 60 * 60 * 1000L;
+
+    /** A time today, rounded down to the previous minute */
+    private static long MODIFIED_TODAY = (System.currentTimeMillis() / 60_000) * 60_000;
```
roxspring, line 454: Throwing away seconds here allows these precise times to be used accurately even on low resolution file systems such as FAT32.

michael-o:
What is the general benefit you expect here? Not changing last modified date?

roxspring:
Yes. That’s the point, as discussed in the jira issue and dev mailing lists. Unnecessarily  modifying the file when the filter content hasn’t changed can lead to wasted work downstream.

michael-o:
Please change the buffer code to: https://github.com/apache/maven-wagon/commit/92c0d2a858a38d9b3d0aefc0f9298d5c31c7e508

roxspring:
@michael-o I think I've fixed the Buffer code as you requested - I'd be grateful of another review!

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -1810,34 +1816,101 @@ public static void copyFile( @Nonnull File from, @Nonnull File to, @Nullable Str
...
-            }
-        }
-        else
-        {
-            if ( to.lastModified() < from.lastModified() || overwrite )
-            {
-                copyFile( from, to );
+                if ( overwrite || !to.exists() )
+                {
+                    try ( Writer fileWriter = new OutputStreamWriter( new FileOutputStream( to ), encoding ) )
```
olamy, line 1845: what about `Files.newBufferedWriter(  )`

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -1810,34 +1816,101 @@ public static void copyFile( @Nonnull File from, @Nonnull File to, @Nullable Str
...
             {
                 encoding = Charset.defaultCharset().name();
             }
-            
+
             // buffer so it isn't reading a byte at a time!
             try ( Reader fileReader =
-                new BufferedReader( new InputStreamReader( new FileInputStream( from ), encoding ) );
-                            Writer fileWriter = new OutputStreamWriter( new FileOutputStream( to ), encoding ) )
+                    new BufferedReader( new InputStreamReader( new FileInputStream( from ), encoding ) ) )
```
olamy, line 1835: what about `Files.newBufferedReader(  )`?

roxspring: Done.

src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java
```diff
@@ -441,9 +449,300 @@ public void copyFile1()
             testFile1.lastModified() == destination.lastModified());*/
     }
 
+    private static long MILLIS_PER_DAY = 24 * 60 * 60 * 1000L;
```
michael-o, line 452: Why not use TimeUnit for this?

roxspring: Happy to use `TimeUnit.DAYS.toMillis(1)` instead. Assumed `ChronoUnit` and friends weren't available but had forgotten `TimeUnit`!

roxspring:
> Does it actually make sense to introduce a threshold? E.g., > 1 MB?

You're suggesting that we apply the new logic only to input >1MB and for smaller files we presumably process them in memory and compare that to the existing written file, right?

In general I dislike maintaining multiple equivalent codepaths and switching between them conditionally as typically one is used far more often and the others languish in bit rot, becoming a maintenance burden. Sometimes this approach is necessary to gain performance in different circumstances, but is that clearly the case here?

I'm not sure it is. For a start I don't remember what rounding errors might have been involved in the 1 vs 8 millisecond comparison from the PR description, but more importantly that's comparing broken to fixed behaviour - the 1 millisecond figure timing is for the existing "unconditionally overwrite when filtering", and we don't yet have comparable numbers for "process them in memory and compare that to the existing written file".

If an additional 7 milliseconds per filtered resource is enough to cause panic then I guess I can look into a thresholded approach but my inclination is to avoid that effort until it proves necessary.

michael-o:
> 
> 
> > Does it actually make sense to introduce a threshold? E.g., > 1 MB?
> 
> You're suggesting that we apply the new logic only to input >1MB and for smaller files we presumably process them in memory and compare that to the existing written file, right?
> 
> In general I dislike maintaining multiple equivalent codepaths and switching between them conditionally as typically one is used far more often and the others languish in bit rot, becoming a maintenance burden. Sometimes this approach is necessary to gain performance in different circumstances, but is that clearly the case here?
> 
> I'm not sure it is. For a start I don't remember what rounding errors might have been involved in the 1 vs 8 millisecond comparison from the PR description, but more importantly that's comparing broken to fixed behaviour - the 1 millisecond figure timing is for the existing "unconditionally overwrite when filtering", and we don't yet have comparable numbers for "process them in memory and compare that to the existing written file".
> 
> If an additional 7 milliseconds per filtered resource is enough to cause panic then I guess I can look into a thresholded approach but my inclination is to avoid that effort until it proves necessary.

Well reasoned, let's leave it as-is.

olamy:
merged via e57180309a64758e161513d80f1185e13dc84826

PR #30, opened 2020-05-16
[MSHARED-860] Deprecate unnecessary Java7Support

roxspring:
I believe that Java 7 has been a minimum requirement for almost a year so here's some overdue cleanup:

- Call `java.nio.file.Files` utility methods directly rather than via reflection
- Move `createSymbolicLink()` utility to `FileUtils`
- Deprecate `Java7Support` and call `java.nio.file.Files` methods directly

MSHARED-860


jira-importer:
Resolve #305

PR #31, opened 2020-05-17
[MSHARED-681] createSymbolicLink() overwrites existing different symlinks

roxspring:
Inspired by @mkarg's https://github.com/apache/maven-shared/pull/12/files but:

- Reimplemented in `FileUtils.createSymbolicLink()` since the utility method was moved from `Java7Support`.
- Read the existing symlink and only delete if it's incorrectly targeted
- Added a test confirming that symlinks are correctly overwritten

### DIFF
--- a/src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java
+++ b/src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java
@@ -1429,7 +1429,7 @@ public class FileUtilsTest
         throws IOException
     {
         // This testcase will pass when running under java7 or higher
-        assumeThat( Os.isFamily(Os.FAMILY_WINDOWS), is(false) );
+        assumeFalse( Os.isFamily( Os.FAMILY_WINDOWS ) );
 
         File file = new File( "src/test/resources/symlinks/src/symDir" );
         assertTrue(FileUtils.isSymbolicLink(file  ));
@@ -1466,7 +1466,8 @@ public class FileUtilsTest
     public void createAndReadSymlink()
         throws Exception
     {
-        assumeThat( System.getProperty( "os.name" ), not( startsWith( "Windows" ) ) );
+        assumeFalse( Os.isFamily( Os.FAMILY_WINDOWS ) );
+
         File file = new File( "target/fzz" );
         FileUtils.createSymbolicLink(  file, new File("../target") );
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java
```diff
@@ -1475,6 +1475,30 @@ public void createAndReadSymlink()
         Files.delete( file.toPath() );
     }
 
+    @Test
+    public void createSymbolicLinkWithDifferentTargetOverwritesSymlink()
+            throws Exception
+    {
+        assumeThat( System.getProperty( "os.name" ), not( startsWith( "Windows" ) ) );
```
elharo, line 1482: I'm a little concerned this might not work (it doesn't in some repos for reasons) so we need to check this in Jenkins to make sure it passes on windows.

roxspring: An understandable concern. Can the Jenkins results be seen somewhere obvious? (it's a shame they're not linked from the PR)

FWIW the condition was copied from the symlink test immediately above, which itself was moved from `Java7SupportTest` only yesterday. Prior to that it appears to have been unchanged here for 3 years: https://github.com/apache/maven-shared-utils/blob/bb2f85e98c3c651ae50b7f642500cb74f50abb0d/src/test/java/org/apache/maven/shared/utils/io/Java7SupportTest.java#L50

On closer inspection it seems that `FileUtilsTest` is already making Windows related assumptions in a variety of ways:

- https://github.com/apache/maven-shared-utils/blob/bb2f85e98c3c651ae50b7f642500cb74f50abb0d/src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java#L446
- https://github.com/apache/maven-shared-utils/blob/92a5ba3c431affcabb2cebd11ca17fe02b15748c/src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java#L1206
- https://github.com/apache/maven-shared-utils/blob/92a5ba3c431affcabb2cebd11ca17fe02b15748c/src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java#L1461
- https://github.com/apache/maven-shared-utils/blob/92a5ba3c431affcabb2cebd11ca17fe02b15748c/src/test/java/org/apache/maven/shared/utils/io/FileUtilsTest.java#L1432

If there's a preferred approach to such assumptions then I'm happy to unify on that - but perhaps in a separate PR?

michael-o: Please use the already given assume approach.

roxspring: Unified explicit calls on `assumeFalse( Os.isFamily( Os.FAMILY_WINDOWS ) );`.

I've left the various path based assumptions as is for now.

olamy:
merged via 8b35ffcf1b410a0870a22ae37f18d96423d42852

jira-importer:
Resolve #206

PR #35, opened 2020-05-27
[MSHARED-893] deprecate file methods we don't need in Java 7+

elharo:
@rfscholte

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
@@ -2046,13 +2046,6 @@ public class FileUtils
     /**
      * Checks whether a given file is a symbolic link.
      *
-     * This only works reliably on java7 and higher. For earlier version we use a highly crappy heuristic
-     * that mostly does not work.
-     * <p>
-     * It doesn't really test for symbolic links but whether the canonical and absolute paths of the file are identical
-     * - this may lead to false positives on some platforms.
-     * </p>
-     *
      * @param file the file to check
      * @throws IOException in case of failure.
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -161,47 +161,52 @@ protected FileUtils()
...
      * Matches the equally named unix command.
      *
-     * @param filename the file path
+     * @param path the file path
      * @return the directory portion excluding the ending file separator
+     * @deprecated use {@code Paths.get(path).getParent().getName()}
      */
-    @Nonnull public static String dirname( @Nonnull String filename )
+    @Deprecated
+    @Nonnull public static String dirname( @Nonnull String path )
```
michael-o, line 169: This hunk is especially great because from a filename you cannot derive a dirname.

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -161,47 +161,52 @@ protected FileUtils()
      * Returns the directory path portion of a file specification string.
      * Matches the equally named unix command.
      *
-     * @param filename the file path
+     * @param path the file path
      * @return the directory portion excluding the ending file separator
+     * @deprecated use {@code Paths.get(path).getParent().getName()}
```
michael-o, line 166: Does this advise also work when the path is relative? Haven't tested.

elharo: It should.

jira-importer:
Resolve #288

PR #36, opened 2020-05-27
[MSHARED-893] deprecate more file methods we don't need in Java 7+

elharo:
@michael-o

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
@@ -792,7 +792,8 @@ public class FileUtils
      * @throws IOException  if <code>source</code> does not exist, <code>destination</code> cannot be
      *                      written to, or an IO error occurs during copying
      * @throws java.io.FileNotFoundException if <code>destination</code> is a directory
-     * @deprecated use {@code org.apache.commons.io.FileUtils.copyFile()}
+     * @deprecated use {@code java.nio.Files.copy(source.toPath(), destination.toPath(), CopyOptions.NOFOLLOW_LINKS,
+     *     CopyOptions.REPLACE_EXISTING)}
      */
     @Deprecated
     public static void copyFile( @Nonnull final File source, @Nonnull final File destination )
@@ -901,6 +902,8 @@ public class FileUtils
      *                     <li><code>destination</code> cannot be written to</li>
      *                     <li>an IO error occurs during copying</li>
      *                     </ul>
+     * @deprecated use {@code java.nio.Files.copy(source.openStream(), destination.toPath(),
+     *     CopyOptions.REPLACE_EXISTING)}
      */
     public static void copyURLToFile( @Nonnull final URL source, @Nonnull final File destination )
         throws IOException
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -573,10 +575,10 @@ private static boolean isValidFile( @Nonnull String file, @Nonnull String... ext
...
     /**
-     * Simple way to make a directory
+     * Simple way to make a directory.
      *
      * @param dir the directory to create
-     * @throws IllegalArgumentException if the dir contains illegal Windows characters under Windows OS.
+     * @throws IllegalArgumentException if the dir contains illegal Windows characters under Windows OS
      * @see #INVALID_CHARACTERS_FOR_WINDOWS_FILE_NAME
      */
     public static void mkdir( @Nonnull String dir )
```
michael-o, line 584: Why didn't you deprecate this one?

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -784,7 +792,9 @@ private static void copyFileToDirectoryIfModified( @Nonnull final File source,
      * @throws IOException  if <code>source</code> does not exist, <code>destination</code> cannot be
      *                      written to, or an IO error occurs during copying
      * @throws java.io.FileNotFoundException if <code>destination</code> is a directory
+     * @deprecated use {@code org.apache.commons.io.FileUtils.copyFile()}
```
michael-o, line 795: What about `Files#copy(Path, Path, Option...)`.

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -913,7 +923,9 @@ public static void copyURLToFile( @Nonnull final URL source, @Nonnull final File
      *                     <li><code>destination</code> cannot be written to</li>
      *                     <li>an IO error occurs during copying</li>
      *                     </ul>
+     * @deprecated use {@code org.apache.commons.io.FileUtils.copyInputStreamToFile()}
```
michael-o, line 926: Available in `Files` too.

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
elharo: This one has some weird logic for handling file names on Windows I'm not sure is available elsewhere.

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
elharo: good idea, done

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
elharo: done

src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
michael-o: Won't Files#createDirectory()` properly fail in such a case? If so, I would rather rely on this one.

elharo: OK

jira-importer:
Resolve #288

PR #37, opened 2020-05-28
[MSHARED-893] deprecate more methods and clean up api doc and tests

elharo:
@michael-o I think this closes out all the FileUtils methods that can be straight-forwardly replaced in Java 7+.

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
@@ -2117,10 +2117,14 @@ public class FileUtils
     }
 
     /**
+     * Create a new symbolic link, possibly replacing an existing symbolic link.
+     * 
      * @param symlink the link name
      * @param target the target
      * @return the linked file
      * @throws IOException in case of an error
+     * @see {@code java.nio.file.Files.createSymbolicLink(Path)} which creates a new
+     *         symbolic link but does not replace exsiting symbolic links
      */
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
```diff
@@ -2103,10 +2117,10 @@ public static boolean isSymbolicLinkForSure( @Nonnull final File file )
...
-     * @param symlink The link name.
-     * @param target The target.
-     * @return The linked file.
-     * @throws IOException in case of an error.
+     * @param symlink the link name
+     * @param target the target
+     * @return the linked file
+     * @throws IOException in case of an error
      */
     @Nonnull public static File createSymbolicLink( @Nonnull File symlink,  @Nonnull File target )
```
michael-o, line 2125: What about `java.nio.file.Files.createSymbolicLink(Path, Path, FileAttribute<?>...)`?

elharo: Perhaps. This method is slightly different, particularly in what happens when the symbolic link already exists. It wraps the `java.nio.file.Files.createSymbolicLink` method for reasons that aren't completely clear to me.

michael-o: OK, I would at least add a `@see` or deprecate and indicate the different behavior.

elharo: done

PR #38, opened 2020-05-28
[MSHARED-894] deprecate close methods since we require Java 7+

elharo:
@michael-o

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/io/IOUtil.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/IOUtil.java
@@ -43,41 +43,36 @@ import java.nio.channels.Channel;
  * copying between sources (<code>InputStream</code>, <code>Reader</code>, <code>String</code> and
  * <code>byte[]</code>) and destinations (<code>OutputStream</code>, <code>Writer</code>,
  * <code>String</code> and <code>byte[]</code>).
- * </p>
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/shared/utils/io/IOUtil.java
```diff
@@ -43,41 +43,36 @@
...
- * <p/>
+ * 
  * <p>Imagine that an InputStream's read() is a very expensive operation, which would usually suggest
  * wrapping in a BufferedInputStream. The BufferedInputStream works by issuing infrequent
  * {@link java.io.InputStream#read(byte[] b, int off, int len)} requests on the underlying InputStream, to
  * fill an internal buffer, from which further <code>read</code> requests can inexpensively get
  * their data (until the buffer runs out).</p>
  * <p>However, the <code>copy</code> methods do the same thing, keeping an internal buffer,
- * populated by {@link InputStream#read(byte[] b, int off, int len)} requests. Having two buffers
+ * populated by {@link InputStream#read(byte[] b, int offset, int len)} requests. Having two buffers
```
michael-o, line 75: The original code refers to it as `off`.

elharo: Fair enough. It's a link so reverted

PR #41, opened 2020-05-29
[MSHARED-860] deprecate PropertyUtils constructor and clean up docs

elharo:
@michael-o


src/main/java/org/apache/maven/shared/utils/PropertyUtils.java
```diff
@@ -89,22 +89,21 @@ public static Properties loadProperties( @Nonnull File file )
...
     public static Properties loadProperties( @Nullable InputStream is )
     {
         try
         {
-            // to make this the same behaviour as the others we should really return null on any error
             Properties result = new Properties();
             if ( is != null )
             {
-                try
+                try ( InputStream in = is )
```
michael-o, line 104: I don't agree with this because have not opened this stream and it is not our task to close it. The one who opens the stream must close it.

elharo: That makes sense. It is a behavior change though. Previously the stream was closed  in lines 120-123. This PR just changes how it's closed to avoid a now deprecated class. If we stop closing it here (and in other methods) then that might be worth a separate bug and PR that shows up in the release notes. What do you think?

michael-o: OK, leave as is and file a bug about the incorrect behavior. This should then be fixed in a subsequent release. Allso add documenation here, that behavior is wrong.

elharo: done: https://issues.apache.org/jira/browse/MSHARED-895

roxspring:
> Question: Doesn't
> 
> ```
> try ( InputStream in = is )
> ```
> 
> end with closing the InputStream in (aka is)?

This has already been discussed here:
https://github.com/apache/maven-shared-utils/pull/41/files#r432415800

In short: Yes it closes the `InputStream` but the original code did too and changing the behaviour would be bad.

jira-importer:
Resolve #305

PR #42, opened 2020-05-29
[MSHARED-860] deprecate constants that are now available in Java 7+

elharo:
@michael-o also updated api doc


jira-importer:
Resolve #305

PR #45, opened 2020-05-29
[MSHARED-898]  prefer JDK classes to DirectoryScanner

elharo:
@michael-o


olamy:
I wonder if it's faster? (jmh tests?)

tisonkun:
@elharo do we have a migration example? It seems that `DirectoryScanner` is still far more fruitful then bare `java.nio.file.DirectoryStream`.

PR #58, opened 2020-07-20
remove call to deprecated method

elharo:
@dennisl


src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomBuilder.java
```diff
@@ -95,10 +94,10 @@ public static Xpp3Dom build( @WillClose InputStream is, @Nonnull String encoding
      * @return the built dom
      * @throws XmlPullParserException in case of an error
      */
-    public static Xpp3Dom build( @WillClose Reader reader, boolean trim )
+    public static Xpp3Dom build( @WillClose Reader in, boolean trim )
```
pzygielo, line 97: @param not updated

PR #60, opened 2020-07-24
[MSHARED-938]

nywitness:
add charset config

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
@@ -280,11 +280,11 @@ public abstract class CommandLineUtils
                         inputFeeder.start();
                     }
 
-                    outputPumper = new StreamPumper( p.getInputStream(), systemOut );
+                    outputPumper = new StreamPumper( p.getInputStream(), systemOut , streamCharset );
                     outputPumper.setName( "StreamPumper-systemOut" );
                     outputPumper.start();
 
-                    errorPumper = new StreamPumper( p.getErrorStream(), systemErr );
+                    errorPumper = new StreamPumper( p.getErrorStream(), systemErr , streamCharset );
                     errorPumper.setName( "StreamPumper-systemErr" );
                     errorPumper.start();
 
--- a/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
+++ b/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
@@ -174,7 +174,7 @@ public class CommandLineUtilsTest
     public void testChineseEncodingIssue()
         throws Exception
     {
-        Commandline commandline = new Commandline( "ping www.baidu.com" );
+        Commandline commandline = new Commandline( "echo 金色传说" );
         StreamConsumer err = new StreamConsumer() {
             @Override
             public void consumeLine( String line ) {
--- a/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
+++ b/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
@@ -24,6 +24,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 
+import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Locale;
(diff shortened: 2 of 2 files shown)


elharo (commented):
Is it possible to test this fix?

nywitness:
Would it be helpful to add a test case in org.apache.maven.shared.utils.cli.CommandLineUtilsTest?
I'm wondering if you can find the difference in a non-chinese os.
I will paste a test screenshot to clarify the difference. Hope it helps.

nywitness:
image

elharo (commented):
Yes, a unit test is helpful, especially when fixing a bug. It should fail without the PR and pass with this PR. That was we know the bug is actually fixed.

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -168,4 +169,25 @@ private void assertCmdLineArgs( final String[] expected, final String cmdLine )
         assertEquals( expected.length, actual.length );
         assertEquals( Arrays.asList( expected ), Arrays.asList( actual ) );
     }
+
+    @Test
+    public void testChineseEncodingIssue()
+        throws Exception
+    {
+        Commandline commandline = new Commandline( "ping www.baidu.com" );
```
michael-o, line 177: No, we cannot have tests which require outbound access. Espcially ICMP. Please change test.

michael-o (changes requested):

src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
```diff
@@ -280,11 +280,11 @@ public Integer call()
                         inputFeeder.start();
                     }
 
-                    outputPumper = new StreamPumper( p.getInputStream(), systemOut );
+                    outputPumper = new StreamPumper( p.getInputStream(), systemOut , streamCharset );
```
michael-o, line 283: Space before comma is not required.

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -168,4 +169,25 @@ private void assertCmdLineArgs( final String[] expected, final String cmdLine )
         assertEquals( expected.length, actual.length );
         assertEquals( Arrays.asList( expected ), Arrays.asList( actual ) );
     }
+
+    @Test
+    public void testChineseEncodingIssue()
```
michael-o, line 174: This test is completely pointless:

1. The `echo` will use the encoding supplied the entire system. There is no guarantee that GBK is the system encoding.
2. You never verify the output of th command to be what you expect.

What you require is an application that wroduces GBK *bytes* , those are read by Java with GBK into a `String` then you need to compared this value.

michael-o (changes requested):

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
nywitness: As far as i am concerned, the parameter `commandLine` of `CommandLineUtils.executeCommandLineAsCallable` is used to create a `Process` object, which is a result of  `Runtime.getRuntime().execute()`. This `execute()` method uses different encoding depending on different system. 

Any idea of producing GBK bytes using `CommandLineUtils.executeCommandLineAsCallable`? Or i can modify the test to use system encoding rather than using GBK.

nywitness: By the way, this test will pass without this PR if system encoding is the same as the result of `Character.defaultCharset()`. So the influence of this fix may not be very obvious.

michael-o: > As far as i am concerned, the parameter commandLine of CommandLineUtils.executeCommandLineAsCallable is used to create a Process object, which is a result of Runtime.getRuntime().execute(). This execute() method uses different encoding depending on different system.

Why do you think so? It uses the same encoding as the surrounding Java process does. You cannot change this really on Windows, on Unix you can pass LC_ALL to the env.

> Any idea of producing GBK bytes using CommandLineUtils.executeCommandLineAsCallable? Or i can modify the test to use system encoding rather than using GBK.

You have two options:

1. Modify `file.encoding` and set back in the finally block. Implies you read the output stream. I don't exactly know whether tests can run in parallel in the same JVM, this could break other tests.
2. Write a simple Java program, put it in `src/test/java`, call the `.class` file with Java from within the test. It should use `System.out` as a byte-oriented stream which will write bytes according to GBK. Read those with the consumer and check when normalized back to UTF-16.

nywitness: > Write a simple Java program, put it in src/test/java, call the .class file with Java from within the test. It should use System.out as a byte-oriented stream which will write bytes according to GBK. Read those with the consumer and check when normalized back to UTF-16.

I tried producing gbk bytes with `System.out.println(new String("金色传说".getBytes(), "GBK")`. When comparing value in the cousumer, test passes on a windows-gbk platform but fails on a mac-utf-8 platform. Maybe it's not right to produce gbk bytes like that.

michael-o: That's wrong. You need to use System.out as an byte stream, not a char stream: `byte[] bytes = "...".getBytes(encoding)` then `System.out.write(bytes)`.

jira-importer:
Resolve #302

PR #61, opened 2020-08-13
[MSHARED-951] fix exception handling

elharo:
@pzygielo

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomBuilder.java
+++ b/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomBuilder.java
@@ -46,8 +46,8 @@ public class Xpp3DomBuilder
 
     /**
      * @param reader {@link Reader}
-     * @return the built dom.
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomBuilder.java
```diff
@@ -84,14 +84,14 @@ public static Xpp3Dom build( @WillClose InputStream is, @Nonnull String encoding
         }
         catch ( UnsupportedEncodingException e )
         {
-            throw new RuntimeException( e );
+            throw new XmlPullParserException( e );
         }
     }
 
     /**
      * @param reader {@link Reader}
```
pzygielo, line 92: There is no such parameter `reader`, as in #58 it was renamed to `in`.

elharo: fixed

PR #62, opened 2020-08-14
[MSHARED-952] make PrettyPrintXmlWriter platform independent

elharo:


michael-o:
Isn't the test portable already when the proper line separator is used throughout?

elharo:
It passes on all platforms ATM but it's not testing the same thing on all platforms. E.g. if we had a bug in handling \r\n then the bug wouldn't be seen when running on Linux or the Mac. Or if we had a bug handing \n it wouldn't be found on Windows.

michael-o:
But the purpose of the test is to test the tags and not the line separators? You have removed `unifyLineSeparators()` which makes it better.

elharo:
The general contract of a unit test is running a function with known, fixed input and comparing it to known fixed output. `unifyLineSeparators()` means the input and output to which its applied are no longer fixed. They're platform dependent. While we do occasionally need tests for specific issues that crop up on this or that platform, there's no need for it here.

michael-o:
Agreed.

elharo:
Fixing this has reveled some more platform dependent code in our output code which interferes with reproducible builds.

PR #68, opened 2020-12-29
[MSHARED-969] Environment variable with null value

slawekjaranowski:


src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -168,4 +169,18 @@ private void assertCmdLineArgs( final String[] expected, final String cmdLine )
         assertEquals( expected.length, actual.length );
         assertEquals( Arrays.asList( expected ), Arrays.asList( actual ) );
     }
+
+    @Test
+    public void environmentVariableWithNullShouldBeAddToList() {
```
pzygielo, line 174: perhaps `Added`

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -168,4 +169,18 @@ private void assertCmdLineArgs( final String[] expected, final String cmdLine )
...
+
+        Commandline commandline = new Commandline();
+        commandline.addEnvironment("TEST_NULL_ENV", null);
+
+        String[] environmentVariables = commandline.getEnvironmentVariables();
+        assertNotNull(environmentVariables);
+        for (String variable: environmentVariables) {
+            assertFalse("Variable '" + variable + "' shouldn't be set",
```
pzygielo, line 182: conflicts with test name

slawekjaranowski:
@pzygielo thanks for review, I did fix

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
```diff
@@ -241,15 +241,16 @@ public void addSystemEnvironment()
     public String[] getEnvironmentVariables()
     {
         addSystemEnvironment();
-        String[] environmentVars = new String[envVars.size()];
-        int i = 0;
+        List<String> environmentVars = new ArrayList<>();
         for ( String name : envVars.keySet() )
```
mkarg, line 245: While I think the solution is working, I wonder whether it solves the actual cause: At least on Windows, it is impossible to stored environment variables with a `NULL` value. Try this effectly removes it from the environment, according to the Windows API docs. So at least on Windows the correct solution would be that this variable name-value-pair must not be put into `envVars` right from the start. In turn, *here* is no change to be made.

Can someone conform that it is valid on other OS to actually have an environment variable in the environment that stores `NULL` (not the String but `null` in the Java sense)?

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
```diff
@@ -241,15 +241,16 @@ public void addSystemEnvironment()
...
             String value = envVars.get( name );
-            environmentVars[i] = name + "=" + value;
-            i++;
+            if ( value != null )
+            {
+                environmentVars.add( name + "=" + value );
+            }
         }
-        return environmentVars;
+        return environmentVars.toArray( new String[0] );
```
mkarg, line 253: We could speed up this code a very little bit (some nanoseconds though) if we use `new String[environmentVars.size()]` instead. It prevents dropping the empty array and creating a new one, hence reduces GC pressure.

pzygielo: > We could speed up this code a very little bit (some nanoseconds though) if we use `new String[environmentVars.size()]` instead. It prevents dropping the empty array and creating a new one, hence reduces GC pressure.

It seems, even across Apache repositories (https://github.com/apache/kafka/pull/9750, https://github.com/apache/tika/pull/372), that using empty array exactly is popular now.

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
slawekjaranowski: reply on jira issue - here we don't set system environment it happen later

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
slawekjaranowski: From InteliJ inspection:

> In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.

mkarg (commented):
I did not talk about popularity neither about reflection, but solely about creating garbage needlessly. I doubt the race condition actually will happen here, as the variable is effectively final at that point.

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -168,4 +169,18 @@ private void assertCmdLineArgs( final String[] expected, final String cmdLine )
         assertEquals( expected.length, actual.length );
         assertEquals( Arrays.asList( expected ), Arrays.asList( actual ) );
     }
+
+    @Test
+    public void environmentVariableWithNullShouldNotBeAddedToList() {
+
+        Commandline commandline = new Commandline();
+        commandline.addEnvironment("TEST_NULL_ENV", null);
```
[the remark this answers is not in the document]
slawekjaranowski: As I wrote before Command Line object has local map of environment variable which are provided to Runtime.exec after translate this map to list of `name=value`. Translation is done in getEnvironmentVariables.

Please look at my last comment at jira and new test in last update.

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
elharo: I agree this is a real bug. Rather than fixing it here, perhaps `addEnvironmentVariable` should throw a NullPointerException for null values?

slawekjaranowski: Throwing exception in this case will has impact on caller, eg has impact on https://issues.apache.org/jira/browse/MINVOKER-273

In my proposition we have new feature - when environment variable has null value will be removed from executing job environment.

Look at unit tests.

Different approach can be  throw exception for null values and add next method to possibility remove/exclude specific environment variables.

MartinKanters (approved):
I agree with @slawekjaranowski and @elharo, this does fix potential bugs. I approve the changes. Thanks @slawekjaranowski!

jira-importer:
Resolve #282

PR #70, opened 2021-01-21
[MNG-6915] Add a helper method to get the terminal width

gnodet:


src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -201,4 +201,21 @@ public void run()
             Runtime.getRuntime().addShutdownHook( shutdownHook );
         }
     }
+
+    /**
+     * Get the terminal width or 0 if the width can not be determined.
```
elharo, line 206: nit: can not --> cannot

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -201,4 +201,21 @@ public void run()
...
+     *
+     * @return the terminal width
+     */
+    public static int getTerminalWidth()
```
elharo, line 210: This needs a test.

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -201,4 +201,21 @@ public void run()
...
+    {
+        if ( JANSI )
+        {
+            return AnsiConsole.getTerminalWidth();
+        }
+        else
+        {
+            return 0;
```
elharo, line 218: -1 is a more common "I couldn't figure out the correct value" flag in these sorts of methods.

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
gnodet: There's no way to easily ensure the test is run from inside a terminal. This would always fail when running from an IDE for example.

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
gnodet: Up to you, jansi returns 0 in that case, but I can check and translate it to -1.

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
gnodet: I've added a unit test, but I'm a bit skeptical as it's really tied to Jansi's internal and is mostly about checking that `getTerminalWidth()` returns the value that will be returned by the underlying implementation.

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
gnodet: Fixed.  The related maven PR does not have to be changed since the test is `<= 0`
https://github.com/apache/maven/pull/434/commits/1cf1487af8081faa55eb8bd9956c8f182f76b4aa#diff-ff39a40ae9303293688fbd2f79ad043c73816772bbd153e51681fdb8c384d085R174-R175

pom.xml
```diff
@@ -70,7 +70,7 @@
     <dependency>
       <groupId>org.fusesource.jansi</groupId>
       <artifactId>jansi</artifactId>
-      <version>2.0.1</version>
+      <version>2.2.0</version>
       <optional>true</optional>
```
elharo, line 74: Why is this marked optional instead of simply declaring it? It feels like unnecessary complexity. The model code checks for its presence but the test code doesn't seem to.

rfscholte: It must be optional, it is Maven that's responsible to provide it and to decide which features are supported.
Maven 3.3.9 and earlier would otherwise have mixed output.

MartinKanters (dismissed):
Looks good to me, @gnodet! Can you please rebase on master and squash your commits?

gnodet:
@MartinKanters rebased and squashed

elharo:
running through Jenkins now: https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven-shared-utils/job/MNG-6915/

PR #74, opened 2021-01-22
Windows 9X is long dead

elharo:
@slawekjaranowski

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
+++ b/src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
@@ -34,7 +34,6 @@ import org.apache.maven.shared.utils.Os;
 import org.apache.maven.shared.utils.StringUtils;
 import org.apache.maven.shared.utils.cli.shell.BourneShell;
 import org.apache.maven.shared.utils.cli.shell.CmdShell;
-import org.apache.maven.shared.utils.cli.shell.CommandShell;
 import org.apache.maven.shared.utils.cli.shell.Shell;
 
 /**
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
Maybe `org.apache.maven.shared.utils.cli.shell.CommandShell` should be also removed - I don't see usage of it in this project.

michael-o:
I agree with @slawekjaranowski

elharo:
Eventually, yes. However it's public API so I don't want to remove it until it's been deprecated for a while longer and ideally we're ready for a major version bump.

slawekjaranowski:
I afraid that this never will happened developers don't look for deprecated ...
The same result will be remove it and release next major version.

Looking for this class on GitHub I found 4 repositories where it is used,  one of them is archived, two don't have any releases so impact will not be huge.

elharo:
I'm not comfortable removing public API without reasonable notice. Sounds like there are existing uses.

slawekjaranowski:
So please deprecate class with reason. 

Another case is that apache components don't have any release plan / calendar so nobody know when next release will be placed.
Also there is no policy about clearing deprecated items.

michael-o:
@elharo This won't even run in Windows 9x because there isn't Java 7 for.

elharo:
That still doesn't mean we should break the build of projects that have this in their tree. Deprecate and warn them. We can take it out in a few releases.

slachiewicz:
`CmdShell` was marked as Deprecated with MSHARED-860 in https://github.com/apache/maven-shared-utils/commit/cb448b90918a5e894c38b86d00d402638f195502 and released with 3.3.3 version.

MartinKanters:
@elharo I guess we can merge this then, right? Nobody had any objections against your code, apart from that we had to deprecate the classes, which @slachiewicz pointed out was done last year.

slachiewicz (approved):

PR #67, opened 2020-12-02
[MNG-6380] Option -Dstyle.color=always doesn't force color output

gnodet:

PR #77, opened 2021-01-22
replace deprecated fields

elharo:
@gnodet


gnodet:
I think the commit at https://github.com/apache/maven-shared-utils/pull/67/commits/09e880253655b05bc63252f57157043cf5d87868 kinda removes the need for this change.

PR #78, opened 2021-01-23
remove unused plexus dependency

elharo:
@michael-o

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,12 +113,6 @@
     </dependency>
     <dependency>
       <groupId>org.codehaus.plexus</groupId>
-      <artifactId>plexus-container-default</artifactId>
-      <version>2.1.0</version>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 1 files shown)


MartinKanters:
@elharo Is there any specific reason @michael-o should look at this? IMO I would say that if the code still compiles without this dependency, we should be good in general?

elharo:
Anyone can review. I try to request reviews roughly round robin amongst active committers unless I believe there's a particular reason someone might be interested in a PR.

slachiewicz (approved):

slachiewicz:
Build failed:

------

[ERROR] /home/jenkins/workspace/en-box_maven-shared-utils_master@2/m/src/main/java/org/apache/maven/shared/utils/cli/javatool/AbstractJavaTool.java:[28,35] package org.codehaus.plexus.logging does not exist

MartinKanters:
@slachiewicz @elharo looks like we could use GitHub Actions in this repo.. :) 
Imo we should revert this PR, do you guys have any concern with that?

PR #80, opened 2021-02-14
Test, fix, and deprecate buggy copyDirectory methods

elharo:
@michael-o 
I found this when I unignored an ignored test. And then I made the mistake of pulling on that string. 
Unignore copyDirectory tests and fix broken behavior that caused them to be ignored. Deprecate these methods in favor of better maintained and documented Apache Commons IO methods from which they were copied and pasted 10+ years ago.


michael-o (approved):

jira-importer:
Resolve #284

PR #81, opened 2021-03-07
Best practices = https://

I-Cat:
Https:// follow best practices


michael-o:
The entire license is opaque and cannot be changed unless a new version is released.

PR #82, opened 2021-03-07
Up date https and added

I-Cat:
# updated http to https and added 
+ "Please"
# Consider adding
+ Licence(s) 
If more than one licence is present

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,23 +1,23 @@
 <?xml version='1.0' encoding='UTF-8'?>
 <!--
   Licensed to the Apache Software Foundation (ASF) under one
-  or more contributor license agreements. See the NOTICE file
+  or more contributor license agreements.Please see the NOTICE file
   distributed with this work for additional information
   regarding copyright ownership. The ASF licenses this file
   to you under the Apache License, Version 2.0 (the
   "License"); you may not use this file except in compliance
-  with the License. You may obtain a copy of the License at
+  with the License. You can obtain a copy of the License at the following link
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1,23 +1,23 @@
...
 
-  http://www.apache.org/licenses/LICENSE-2.0
+  https://www.apache.org/licenses/LICENSE-2.0
 
   Unless required by applicable law or agreed to in writing,
-  software distributed under the License is distributed on an
+  the software distributed under the License is distributed on an
   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-  KIND, either express or implied. See the License for the
+  KIND, either express or implied. Please see the License for the
```
hboutemy, line 16: no you cannot change the license text like this, even if you think your wording is better

pom.xml
```diff
@@ -1,23 +1,23 @@
...
   specific language governing permissions and limitations
   under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
hboutemy, line 20: everything you changed are identifiers, then not to change (and not really used as url to download anything)

the only url that is used to download something (= the xsd file) is the only location you did not change: this one can be changed

hboutemy:
sorry, closed as invalid, as per comments

PR #85, opened 2021-04-23
Javadoc cleanup in preparation of release

mthmulders:


pzygielo:
Forgive me, but is there a chance you would consider removing one `;` from https://github.com/apache/maven-shared-utils/blob/4e93c5621ef45c7cf6a03d96ca7a3c45ac4e2062/src/test/java/org/apache/maven/shared/utils/xml/PrettyPrintXmlWriterTest.java#L37 as well, please? It has no impact on javadoc, but I can't imagine better PR like this :grin:

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -104,7 +104,7 @@ private static void doSystemUninstall()
 
     /**
      * Enables message color (if JAnsi is available).
```
MartinKanters, line 106: Let's call it "Jansi" as it's done in [their source](https://github.com/fusesource/jansi)

```suggestion
     * Enables message color (if Jansi is available).
```

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -104,7 +104,7 @@ private static void doSystemUninstall()
...
-     * @param flag
+     * @param flag To enable JANSI
```
MartinKanters, line 107:
```suggestion
     * @param flag to enable Jansi
```

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -145,6 +146,7 @@ public static MessageBuilder buffer()
 
     /**
      * Create a message buffer with defined String builder.
+     * @param builder Initial content of the message buffer
```
MartinKanters, line 149:
```suggestion
     * @param builder initial content of the message buffer
```

MartinKanters (changes requested):

mthmulders:
> Forgive me, but is there a chance you would consider removing one `;` from
> 
> https://github.com/apache/maven-shared-utils/blob/4e93c5621ef45c7cf6a03d96ca7a3c45ac4e2062/src/test/java/org/apache/maven/shared/utils/xml/PrettyPrintXmlWriterTest.java#L37
> as well, please? It has no impact on javadoc, but I can't imagine better PR like this 😁

Done.

mthmulders:
> Did you remove the default messages on these methods on purpose? If so, don't you think it adds value?

The remarks like "By default, bold cyan" refer to one particular implementation (the `AnsiMessageBuilder`) but they may not apply to other implementations (such as the `PlainMessageBuilder`). The interface should refrain from describing such implementation details, I would argue.

MartinKanters:
> > Did you remove the default messages on these methods on purpose? If so, don't you think it adds value?
> 
> The remarks like "By default, bold cyan" refer to one particular implementation (the `AnsiMessageBuilder`) but they may not apply to other implementations (such as the `PlainMessageBuilder`). The interface should refrain from describing such implementation details, I would argue.

Right, makes sense to me!

MartinKanters (approved):
LGTM, feel free to squash and merge

PR #86, opened 2021-04-26
avoid reusing fixtures

elharo:
@mthmulders

### DIFF
--- a/src/test/java/org/apache/maven/shared/utils/xml/XmlWriterUtilTest.java
+++ b/src/test/java/org/apache/maven/shared/utils/xml/XmlWriterUtilTest.java
@@ -20,6 +20,7 @@ package org.apache.maven.shared.utils.xml;
  */
 
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/maven/shared/utils/xml/XmlWriterUtilTest.java
```diff
@@ -86,7 +75,6 @@ public void testWriteLineBreakXMLWriterInt()
     {
         XmlWriterUtil.writeLineBreak( xmlWriter, 10 );
         writer.close();
-        assertTrue( StringUtils.countMatches( output.toString(), XmlWriterUtil.LS ) == 10 );
```
mthmulders, line 89: Did you remove the assertion on purpose? If so, why?

elharo: nope, just an error. I'll fix it.

elharo: done

mthmulders (approved):
Looks good to me.

PR #93, opened 2021-07-07
[MSHARED-993] Make uninstall resilient to Jansi errors

guylabs:
See https://issues.apache.org/jira/browse/MSHARED-993 for details.

There is no unit/integration test for this as it would require to use Powermock or similar to mock Jansi. If requested, I can add this but I thought that this would be an overkill.

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
@@ -98,7 +98,14 @@ public class MessageUtils
     {
         if ( JANSI )
         {
-            AnsiConsole.systemUninstall();
+            try
+            {
+                AnsiConsole.systemUninstall();
+            }
+            catch ( Throwable ex )
+            {
+                // make uninstall resilient to any errors thrown by Jansi
+            }
         }
     }
 
--- a/src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
@@ -102,9 +102,9 @@ public class MessageUtils
             {
                 AnsiConsole.systemUninstall();
             }
-            catch ( Throwable ex )
+            catch ( RuntimeException ex )
             {
-                // make uninstall resilient to any errors thrown by Jansi
+                // make uninstall resilient to any runtime exceptions thrown by Jansi
             }
         }
     }


src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -98,7 +98,14 @@ private static void doSystemUninstall()
     {
         if ( JANSI )
         {
-            AnsiConsole.systemUninstall();
+            try
```
rfscholte, line 101: Why is there a try-catch block required here, while it is not there for `AnsiConsole.systemInstall()`.
If this is an issue, it should be fixed in JAnsi (fix it at the root, instead of handling the symptom)

guylabs: This was already proposed here https://issues.apache.org/jira/browse/MNG-7161 but it seemed that the consensus was to go with that approach, which makes `MessageUtils` less dependent on changes on Jansi which caused this error. Fixing the root cause will fix _this_ specific use case, but there may be others which will break too, and as the `systemUninstall` is called as a last step before Maven finishes the build we can assume that it's safe to swallow these errors. I created https://github.com/fusesource/jansi/issues/214 to fix the root cause.

IMO we should not add it to the `systemInstall` as there you'd like to be aware when we can't install the color support and get the exception during runtime, as if we would swallow that error too, it will be harder to debug the root cause without any exception thrown. If you prefer to have the same handling, then I can add this.

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -98,7 +98,14 @@ private static void doSystemUninstall()
...
+            {
+                AnsiConsole.systemUninstall();
+            }
+            catch ( Throwable ex )
```
elharo, line 105: never catch Throwable. See Effective Java for reasons.

elharo (dismissed):

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
guylabs: Thanks! Fixed.

rfscholte (changes requested):
We're still trying to fix this at JAnsi. This should only be accepted if JAnsi explcitly states they won't fix it.

guylabs:
> We're still trying to fix this at JAnsi. This should only be accepted if JAnsi explcitly states they won't fix it.

There are still no reactions to https://github.com/fusesource/jansi/issues/214. Not sure what the timeline is for the proper fix. Do you maybe have any other insights @rfscholte ?

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -98,7 +98,14 @@ private static void doSystemUninstall()
     {
         if ( JANSI )
         {
-            AnsiConsole.systemUninstall();
+            try
+            {
+                AnsiConsole.systemUninstall();
+            }
+            catch ( RuntimeException ex )
```
elharo, line 105: does it throw a specific runtime exception?

guylabs: No.

rfscholte:
> > We're still trying to fix this at JAnsi. This should only be accepted if JAnsi explcitly states they won't fix it.
> 
> There are still no reactions to [fusesource/jansi#214](https://github.com/fusesource/jansi/issues/214). Not sure what the timeline is for the proper fix. Do you maybe have any other insights @rfscholte ?

We're working on it as we have more ways to get direct contact with the maintainer(s). But I agree, it takes more time, must be the summer ;)

guylabs:
@rfscholte Ok to close this pull request as https://github.com/fusesource/jansi/issues/214 has been fixed now ? Should we then also close https://issues.apache.org/jira/browse/MSHARED-993 ?

rfscholte:
Yes, I'll close them both.

PR #94, opened 2022-01-03
[MSHARED-1014] Optionally inherit system environment variables by Commandline

slawekjaranowski:


src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
```diff
@@ -198,14 +200,13 @@ public void addArguments( String... line )
...
     {
-        //envVars.add( name + "=" + value );
         envVars.put( name, value );
     }
 
     /**
      * Add system environment variables.
      */
-    public void addSystemEnvironment()
+    private void addSystemEnvironment()
```
elharo, line 209: This looks like an API breaking change

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
```diff
@@ -377,6 +382,28 @@ public void clearArgs()
         arguments.clear();
     }
 
+    /**
+     * Indicates whether the environment variables of the current process should be propagated to the executing Command.
```
elharo, line 386: should be --> are

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -22,6 +22,7 @@
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasItemInArray;
 import static org.hamcrest.Matchers.not;
+import static org.hamcrest.collection.IsArrayWithSize.emptyArray;
```
elharo, line 25: This was unfamiliar and confusing when I read it below. A static import is not helpful here.

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -195,6 +190,18 @@ public void environmentVariableFromSystemIsCopied() {
...
+    @Test
+    public void environmentVariableFromSystemIsNotCopiedIfInheritedIsFalse() {
+
+        Commandline commandline = new Commandline();
+        commandline.setShellEnvironmentInherited( false );
+
+        String[] environmentVariables = commandline.getEnvironmentVariables();
+
+        assertNotNull(environmentVariables);
+        assertThat(environmentVariables, emptyArray() );
```
elharo, line 202: assertEquals(0, environmentVariables.length)

would be much easier to read here.

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
slawekjaranowski: Yes I know, this method is confusing because it is used internally in `getEnvironmentVariables`.
So when we switch to `4.0.0` I think that can break api in this point.
Now will be setShellEnvironmentInherited for achieve this feature.

slawekjaranowski:
fixed - @elharo please review

src/main/java/org/apache/maven/shared/utils/cli/Commandline.java
```diff
@@ -65,12 +65,14 @@
 public class Commandline
     implements Cloneable
 {
-    private final List<Arg> arguments = new Vector<Arg>();
+    private final List<Arg> arguments = new Vector<>();
```
cstamas, line 68: Is Vector needed here?

michael-o: That's old code. I wouldn't touch it in THIS PR.

slawekjaranowski: Exactly.
I don't want to mixing changes in one PR.

Probably Vector and below `synchronizedMap` can be replaced for not synchronized collections.

`Commandline` should be one time instance not shared by threads.
But it is another issue.

cstamas (approved):

PR #95, opened 2022-03-04
(doc) Throw more descriptive NPEx

pzygielo:
As suggested in [MJAR-286](https://issues.apache.org/jira/browse/MJAR-286):
> If this is a misconfigured pom, it should at least give a proper error message.

Thus replacing JVM-thrown NPEx with the one with message.


pzygielo:
May I ask for review, please?

slawekjaranowski (commented):
simple unit test will be appreciated 😉
it is public method

slawekjaranowski (approved):

PR #96, opened 2022-06-08
[MSHARED-1079] add build() method and document toString() method

kwin:
improve Javadoc

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/logging/MessageBuilder.java
+++ b/src/main/java/org/apache/maven/shared/utils/logging/MessageBuilder.java
@@ -19,8 +19,13 @@ package org.apache.maven.shared.utils.logging;
  * under the License.
  */
 
+import java.util.Formatter;
+
 /**
  * Message builder that supports configurable styling.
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/maven/shared/utils/logging/MessageBuilder.java
```diff
@@ -21,6 +21,8 @@
 
 /**
  * Message builder that supports configurable styling.
+ * An instance of this interface can be retrieved with {@link MessageUtils#buffer()}.
+ * After the message has been build with any of the append methods its content can be retrieved with {@link #build()}.
```
pzygielo, line 25: _build, built, **built**_

kwin: Thanks, replaced by "constructed" in https://github.com/apache/maven-shared-utils/pull/96/commits/35eb1e3a0dd029fb872b153fc04c847089b07377 as "build/built" is kind of overloaded here.

PR #105, opened 2022-07-25
[MSHARED-1107] Make copyFile succeed with source file having

kwin:
lastModified() = 0

Also check for existence of destination file


michael-o:
Is NIO2 prone to this error? If not, why not use the `FileTime` there?

kwin:
NIO2 would throw a proper exception when checking last modified date on a non existing file, but that wouldn't help here either. So an existence check is necessary even when using NIO2 for modified date.

PR #107, opened 2022-07-25
[MSHARED-1109] fix broken IT after previous refactoring

kwin:


kwin:
@michael-o That should fix the broken IT, which regressed due to https://github.com/apache/maven-shared-utils/commit/8822bea7d21d53b106a696fd95f10c494ce20c7a.

michael-o:
Charming...

kwin:
The master build and also this PR cannot be built currently due to another issue with m-resources-p

michael-o:
> The master build and also this PR cannot be built currently due to another issue with m-resources-p

Chicken and egg we'll fix in the next round...

jira-importer:
Resolve #310

PR #112, opened 2022-07-27
[MSHARED-1116] Drop findbugs-maven-plugin from configuration

slawekjaranowski:
Not present in parent, no execution section, inactive


michael-o (approved):
We don't do bugs, let's drop it.

jira-importer:
Resolve #208

PR #113, opened 2022-07-27
[MSHARED-1072] fix blocking in StreamFeeder

slawekjaranowski:
If input stream has no more available data
StreamFeeder was block forever

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/cli/CommandLineTimeOutException.java
+++ b/src/main/java/org/apache/maven/shared/utils/cli/CommandLineTimeOutException.java
@@ -20,16 +20,15 @@ package org.apache.maven.shared.utils.cli;
  */
 
 /**
+ * Report a timeout for executing process.
+ *
  * @author Olivier Lamy
- * 
+ *
  */
 public class CommandLineTimeOutException
     extends CommandLineException
 {
 
-    /**
-     * 
-     */
     private static final long serialVersionUID = 7322428741683224481L;
 
     /**
@@ -41,4 +40,11 @@ public class CommandLineTimeOutException
         super( message, cause );
     }
 
+    /**
+     * @param message The message of the exception.
+     */
+    public CommandLineTimeOutException( String message )
+    {
+        super( message );
+    }
 }
--- a/src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
@@ -27,6 +27,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 import java.util.StringTokenizer;
+import java.util.concurrent.TimeUnit;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -49,7 +50,7 @@ public abstract class CommandLineUtils
         implements StreamConsumer
     {
 
-        private final StringBuffer string = new StringBuffer();
+        private final StringBuilder string = new StringBuilder();
 
         private static final String LS = System.getProperty( "line.separator", "\n" );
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/apache/maven/shared/utils/cli/CommandLineTimeOutException.java
```diff
@@ -41,4 +40,11 @@ public CommandLineTimeOutException( String message, Throwable cause )
         super( message, cause );
     }
 
+    /**
+     * @param message The message of the exception.
+     */
+    public CommandLineTimeOutException( String message )
```
michael-o, line 46: `Timeout` according to https://www.oxfordlearnersdictionaries.com/definition/english/timeout?q=timeout

src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
```diff
@@ -288,31 +279,13 @@ public Integer call()
                     errorPumper.setName( "StreamPumper-systemErr" );
                     errorPumper.start();
 
-                    int returnValue;
-                    if ( timeoutInSeconds <= 0 )
+                    if ( timeoutInSeconds > 0 && !p.waitFor( timeoutInSeconds, TimeUnit.SECONDS ) )
```
michael-o, line 282: Where is the kill for this process? It may continue to run as a zombie forever.

michael-o (changes requested):

src/main/java/org/apache/maven/shared/utils/cli/CommandLineTimeOutException.java
slawekjaranowski: Existing exception name ... I can rename .. but will be break, maybe ok for 4.x 😄

src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
slawekjaranowski: There is defined `ShutDownHook` lines 237 - 253, and it is also called in line 362

There are also `pumper` threads which consume all outputs of process.

src/main/java/org/apache/maven/shared/utils/cli/CommandLineTimeOutException.java
michael-o: No, leave it

famod:
Coming from https://issues.apache.org/jira/browse/MRELEASE-1114, what's the status of this? Are there any blockers?

kwin:
@slawekjaranowski Given that maven-shared-utils is mostly dead, I am wondering whether you can provide the fix instead for https://github.com/codehaus-plexus/plexus-utils/tree/master/src/main/java/org/codehaus/plexus/util/cli and afterwards we convert maven-gpg-plugin to using plexus-utils again instead of maven-shared-utils.

michael-o:
> @slawekjaranowski Given that maven-shared-utils is mostly dead, I am wondering whether you can provide the fix instead for https://github.com/codehaus-plexus/plexus-utils/tree/master/src/main/java/org/codehaus/plexus/util/cli and afterwards we convert maven-gpg-plugin to using plexus-utils again instead of maven-shared-utils.

I think this fix deserves a last patch release and done.

slawekjaranowski:
> > @slawekjaranowski Given that maven-shared-utils is mostly dead, I am wondering whether you can provide the fix instead for https://github.com/codehaus-plexus/plexus-utils/tree/master/src/main/java/org/codehaus/plexus/util/cli and afterwards we convert maven-gpg-plugin to using plexus-utils again instead of maven-shared-utils.
> 
> I think this fix deserves a last patch release and done.

Ok, I will rebase and check again 😄

slawekjaranowski:
@michael-o refresh with current master, conflict resolved
Tomorrow I would to check again.

src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
```diff
@@ -436,7 +387,7 @@ public static String[] translateCommandline(String toProcess) throws CommandLine
         boolean inEscape = false;
         int state = normal;
         final StringTokenizer tok = new StringTokenizer(toProcess, "\"\' \\", true);
```
michael-o, line 389: Off topic: This is sooo ugly...

michael-o (approved):

slawekjaranowski:
Tested with m-release-p ... 

When we want to upgrade version in m-release-p we need #138

PR #123, opened 2022-11-25
[MSHARED-1172] Deprecate redundant isEmptyString method

elharo:
deprecate a string utility method that doesn't belong in this package, wasn't correctly documented, isn't tested, and is duplicated in 72 other places

@slawekjaranowski

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
@@ -148,8 +148,8 @@ public class Xpp3DomUtils
     }
 
     /**
-     * @deprecated use <code>String.isBlank(str)</code> (Java 11+) 
-     *             or <code>str == null || org.apache.commons.lang3.StringUtils.isBlank(str)</code>
+     * @deprecated use <code>str == null || String.isBlank(str)</code> (Java 11+) 
+     *             or <code>org.apache.commons.lang3.StringUtils.isBlank(str)</code>
      * @param str the string to be checked
      * @return <code>true</code> if the string is null, empty, or whitespace only; <code>false</code> otherwise
      */
--- a/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
@@ -148,8 +148,8 @@ public class Xpp3DomUtils
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
```diff
@@ -148,15 +148,15 @@ private static boolean isMergeChildren( Xpp3Dom dominant )
     }
 
     /**
-     * @param str The string to be checked.
-     * @return <code>true</code> in case string is empty <code>false</code> otherwise.
+     * @deprecated use <code>String.isBlank()</code> (Java 11+)
```
michael-o, line 151: `String#isBlank()` is *not* a replacement because it will cause a NPE if null.

michael-o (changes requested):

src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
elharo: fixed

elharo:
PTAl @michael-o

src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
```diff
@@ -148,15 +148,15 @@ private static boolean isMergeChildren( Xpp3Dom dominant )
     }
 
     /**
-     * @param str The string to be checked.
-     * @return <code>true</code> in case string is empty <code>false</code> otherwise.
+     * @deprecated use <code>str == null || String.isBlank(str)</code> (Java 11+)
```
kwin, line 151: Why not recommending `https://github.com/apache/maven-shared-utils/blob/f4383c55f77ad4480f146481bf3192bc52c3af30/src/main/java/org/apache/maven/shared/utils/StringUtils.java#L145` as replacement? Or should we rather deprecate that method as well?

michael-o: This is fine as well

michael-o (approved):

src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomUtils.java
elharo: IMHO

JDK > org,apache.commons > Guava > maven-shared-utils > plexus-utils > random classes  where it doesn't belong

Realistically, this functionality is reinvented so often and so frequently, we'll never get down to a single version. However this case is particularly egregious.

PR #127, opened 2022-11-27
[MSHARED-1169] deprecate join methods now available in Java 8 String class

elharo:
@slachiewicz


michael-o (approved):

slachiewicz (approved):

slachiewicz:
Maybe If we plan to deprecate something would be good to have Jira so it can be included also in release note?

jira-importer:
Resolve #321

PR #133, opened 2023-03-05
suggest java.nio.Files.walkFileTree as replacement for DirectoryScanner

elharo:


michael-o (commented):
Looking at those I don't see both as a 1:1 alternative, no? It still requires addon code.

elharo:
That is correct. It's not a 1:1 replacement, but it does serve the same purpose. Starting from a green field it's no more work to write new code on top of Files.walkFileTree or DirectoryStream instead of DirectoryScanner; and those are more familiar since they're in the JDK. However, migrating from existing DirectoryScanner to one of the JDK alternatives is not a simple search and replace. I'm working on some of those migrations now which is where I noticed Files.walkFileTree was also a good option.

michael-o (changes requested):
Please mention the missing bits in your deprecation notice.

elharo:
Are there missing bits you're especially concerned about like the default exclusions? Nothing is really missing from the alternatives. It's just that you have to write different code, sort of like the difference between going from 14th Street to Penn Station via subway or taxi. Both have the same endpoints but the route and procedure are quite different.

elharo:
https://github.com/apache/maven-dependency-analyzer/pull/81 is an example of the sort of change needed to replace these methods, though in that case it's moving from the plexus-utils version

slachiewicz (approved):

michael-o (approved):

PR #140, opened 2023-05-06
[MSHARED-1072] Poll data from input stream

slawekjaranowski:
Input stream can be a System.in - all read will be blocked 
We need read data in no blocking mode


michael-o:
Does this always apply? Are there cases were in the stream is not `stdin`?

slawekjaranowski:
> Does this always apply? Are there cases were in the stream is not `stdin`?

It will be safer to use non blocking io in such case.
Even when stream will be not System.in it will work.

We also can not close provided stream because can be used for multiple executing of invoker.

michael-o (approved):

jira-importer:
Resolve #313

PR #143, opened 2023-05-16
[MSHARED-1255] use dependencyManagement to set hamcrest-core version

elharo:
Removes dependency warning but still selects the right version of hamcrest-core in transitive dependencies. See https://hamcrest.org/JavaHamcrest/distributables

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,16 @@
     <mavenVersion>3.2.5</mavenVersion>
   </properties>
 
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.hamcrest</groupId>
+        <artifactId>hamcrest-core</artifactId>
+        <version>2.2</version>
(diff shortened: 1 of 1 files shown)


michael-o:
I don't understand this. If a direct dep uses the wrong version then there is the place to fix? Or do we have a conflict in trans deps here?

elharo:
See https://hamcrest.org/JavaHamcrest/distributables for the detailed explanation.

This is a weird case that crops up when using JUnit 4 with Hamcrest 2 because Hamcrest moved classes around between artifacts in version 2. Without this JUnit 4 will pull hamcrest-core 1.4.x into the tree, which contains some of the same classes as org.hamcrest:hamcrest:2.2. This can cause split package issues. In 2.2 hamcrest-core is an empty jar to override this.

michael-o (approved):
Your explanation is reasonable.

PR #154, opened 2023-06-12
[MSHARED-1269] Deprecate maven-shared-utils

gnodet:
JIRA issue: https://issues.apache.org/jira/browse/MSHARED-1269

Maven-shared-utils was created so that plexus-utils could be deprecated on kinda _come back_ into the maven land.  However, this did never happened and we end-up with the exact same library duplicated and duplicate maintenance effort.

The only problematic point is the `org.apache.maven.shared.utils.logging` package which is used by several components and does not exists in plexus-utils.  Maven 4 provides a similar API already with the `MessageBuilderFactory` service which allows the creation of `MessageBuilder`.  I thus propose to get rid completely of maven-shared-utils.

The [maven PR](https://github.com/apache/maven/pull/1158) get rid of maven-shared-utils in maven core.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,12 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 -->
+Deprecation
+======================
+The [Apache Maven Shared Utils](https://maven.apache.org/shared/maven-shared-utils/) library has been deprecated in favor 
+of [Plexus Utils](https://github.com/codehaus.plexus/plexus-utils) and [Plexus XML](https://github.com/codehaus.plexus/plexus-xml).
+Please switch to using those libraries.
+
 Contributing to [Apache Maven Shared Utils](https://maven.apache.org/shared/maven-shared-utils/)
 ======================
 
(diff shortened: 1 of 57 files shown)


elharo (changes requested):
IMHO plexus-utils is the one that needs to go, not this. Plexus is not an apache project, and doesn't have the same controls or support as Maven. The preference order is

JDK > Apache Commons > maben-shared-utils > plexus

elharo:
Where we can get rid of this, do so. There are many cases where there's better in the JDK or Apache commons these days.  But we should not replace this with plexus utils.

elharo:
The more I look at this the more places I find where these two have diverged. E.g. plexus-utils does not seem to have any equivalent of org.apache.maven.shared.utils.logging. Even after a package rename it's not a drop-in replacement. I don't think this proposal is feasible.

slawekjaranowski:
> IMHO plexus-utils is the one that needs to go, not this. Plexus is not an apache project, and doesn't have the same controls or support as Maven. The preference order is
> 
> JDK > Apache Commons > maben-shared-utils > plexus

Here I agree with @elharo, even more many plexus components have an unclear licenses.

PR #158, opened 2023-06-18
[MSHARED-1273] clean up API doc and exceptions

elharo:
still trying to figure out how best to replace these methods completely

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/io/FileUtils.java
@@ -1252,37 +1252,39 @@ public class FileUtils {
     }
 
     /**
-     * Return the files contained in the directory, using inclusion and exclusion Ant patterns,
-     * including the directory name in each of the files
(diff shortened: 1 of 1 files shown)


slachiewicz (approved):

slawekjaranowski (approved):

slawekjaranowski:
Maybe commit message should contain ... `and do not throw IOException` as jira issue
we don't add exceptions

slachiewicz:
so it will be breaking change

elharo:
It isn't throwing IOException now. This PR brings the throws clauses in sync with what the code was already (not) doing.

slachiewicz:
external code that use this methods and catch exception will be broken after update

elharo:
You're right. I think I was mixing up catch blocks for unthrown exceptions and throws clauses for unthrown exceptions in my head. I'll send another PR to clean this up.

slachiewicz:
No problem. You just have to release minor versions and not bugfix

PR #165, opened 2023-07-20
[MSHARED-1289] Deprecate library

kwin:
TODO:
Find replacements for 

1. `org.apache.maven.shared.utils.cli.javatool` and
2. `org.apache.maven.shared.utils.logging`

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/PathTool.java
+++ b/src/main/java/org/apache/maven/shared/utils/PathTool.java
@@ -32,7 +32,9 @@ import java.util.StringTokenizer;
  * in maven-utils of Apache Maven-1.
  * Some external fixes by Apache Committers have been applied later.
  * </p>
+ * @deprecated Use {@code org.codehaus.plexus.util.PathTool} instead.
  */
+@Deprecated
 public class PathTool {
 
     /**
--- a/src/main/java/org/apache/maven/shared/utils/PropertyUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/PropertyUtils.java
@@ -30,7 +30,9 @@ import java.util.Properties;
 
 /**
  * Static utility methods for loading properties.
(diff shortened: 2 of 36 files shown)


elharo (changes requested):
One more time: No. This is exactly wrong. We want to get rid of plexus, not use it more,.

kwin:
Feel free to comment and decline but closing some other committers PR is a no go for me. Also you didn’t state no on the referenced mailing list thread. Reality is that plexus utils got more attention the last years and bugfixes were never backported to maven-shared-utils. We cannot afford maintaining both!

elharo:
We went through this a few weeks ago, This is not new. 

I agree we can't maintain both. Plexus needs to go. That's been decided at least twice and we can't keep relitigating it.

kwin:
Who decided and when? I know the original intent but that failed clearly. Also only 4 committers commented on the aforementioned thread. 3 in favor and you.

elharo:
see https://github.com/apache/maven-shared-utils/pull/154 from less than a month ago.

elharo:
Sorry, I was wrong. It was a month and three days ago; but I still don;t want to have to close this same PR every month. Your time would be better spent replacing existing usages of plexus throughout maven or reviewing unreviewed PRs in this repo, among otehr things.

slachiewicz (approved):

elharo:
Still a hard no on this one. This needs to be dealt with one piece at a time, not as a single PR that deprecates things that are in use and that we don't have replacements for. Even more importantly, we need to eliminate plexus-utils, not reintroduce it. The foggy ownership and licensing story make it infeasible for Apache projects.

slawekjaranowski (changes requested):
I agree with @elharo we should think rather about deprecate plexus-utils especially with licenses issues

So we should deprecate one by one feature which has replacement in JDK or other commons library.

slachiewicz:
Only packages listed by Konrad above must be kept in this lib, or we can find a replacement. All others are just duplicates.

PR #187, opened 2025-02-26
Use caching methods for system properties

elharo:
Minor improvement that avoids repeated lookups of system properties and is a little more direct in code


src/main/java/org/apache/maven/shared/utils/ReaderFactory.java
```diff
@@ -103,7 +104,7 @@ public class ReaderFactory {
      * @deprecated use {@code java.nio.charset.Charset.getDefaultCharset()}
      */
     @Deprecated
-    public static final String FILE_ENCODING = System.getProperty("file.encoding");
```
michael-o, line 106: This is even a non-public property...

PR #190, opened 2025-03-23
windows command default charset is GBK

Chu-wee:


slawekjaranowski:
can you elaborate what problem do you try to resolve?

Chu-wee:
In the Chinese version of the Windows operating system, when the command execution results contain Chinese characters, the output content appears as garbled text.

slawekjaranowski:
We can not set a charset as constant ... what happens when other locale will be used on system?

Chu-wee:
What should I do in this situation? The command is output in GBK, but StreamPumper reads it as garbled text.

PR #194, opened 2025-04-30
Move to org.jline:jansi

slachiewicz:


src/test/java/org/apache/maven/shared/utils/logging/MessageUtilsTest.java
```diff
@@ -78,6 +78,7 @@ public int getTerminalWidth() {
                 false);
         try {
             AnsiConsole.systemInstall();
+            // AnsiConsole.out is not available, need to adjust initialization
```
slachiewicz, line 81: how to convert usage of `AnsiOutputStream` in new lib?

gnodet:
I think the switch is only relevant if Maven 3.x switches to JLine too.

Fwiw, the library still works in Maven 4 for Maven 3.x plugins, and it won't be used for plugins using the Maven 4 API, so I'm not really sure we want to go further.

PR #368, opened 2026-02-04
Prefer StringBuilder for local variables

elharo:
since they don't need synchronization

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/StringUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/StringUtils.java
@@ -1321,7 +1321,7 @@ public class StringUtils {
             if (length == 0) {
                 return "";
             } else {
-                return new StringBuffer(length)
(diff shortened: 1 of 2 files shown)


elharo:
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[FATAL] Non-resolvable parent POM for org.apache.maven.shared:maven-shared-utils:3.4.3-SNAPSHOT: The following artifacts could not be resolved: org.apache.maven.shared:maven-shared-components:pom:47 (absent): Could not transfer artifact org.apache.maven.shared:maven-shared-components:pom:47 from/to central (https://repo.maven.apache.org/maven2): PKIX path validation failed: java.security.cert.CertPathValidatorException: validity check failed and 'parent.relativePath' points at wrong local POM @ line 23, column 11
 @

slachiewicz (approved):

PR #371, opened 2026-03-11
Move off from deprecated library

cstamas:
And move onto org.jline:jansi instead that is actively maintained.

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
@@ -18,9 +18,9 @@
  */
 package org.apache.maven.shared.utils.logging;
 
-import org.fusesource.jansi.Ansi;
-import org.fusesource.jansi.AnsiConsole;
-import org.fusesource.jansi.AnsiMode;
+import org.jline.jansi.Ansi;
+import org.jline.jansi.AnsiConsole;
+import org.jline.jansi.AnsiMode;
 
 /**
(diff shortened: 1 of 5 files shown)


cstamas:
This library remains part of old Maven 3 ecosystem forever. Maven 4 is on JLine already.

elharo:
not sure why we need to keep this for Maven 3?

cstamas:
Because the old library is exported artifact and exported package by Maven Core:
https://github.com/apache/maven/pull/11785/changes#diff-b76b580d9f114f031c237a133170d81d40ead39ddecd91b576f40c2139e31ba0

Hence, doing this sort of a breakage (ie between 3.9.x and 3.10.x) could result that some plugins would work with 3.9.x but not with 3.10.x and other way around, and even worse, may require extra burden for Maven 4 that supports Maven 3 to support this or that or both. 

Best to have 3.10.x "smooth sailing", and just keep this (deprecated) library as dependency, as 4 is coming anyway.

elharo:
Makes sense, or at least why you're closing this makes sense. Can't say I agree with the original decision to export this artifact but that ship sailed long ago. Dependencies == pain.

src/main/java/org/apache/maven/shared/utils/logging/MessageUtils.java
```diff
@@ -44,7 +44,7 @@ public class MessageUtils {
         boolean jansi = true;
         try {
             // Jansi is provided by Maven core since 3.5.0
-            Class.forName("org.fusesource.jansi.Ansi");
```
slachiewicz, line 47: here we should look for both versions - as we may run on old Maven also

PR #377, opened 2026-05-28
build(deps): replace fusesource.jansi with jline.jansi for ANSI support

### LINKED ISSUE #373: Replace org.fusesource.jansi:jansi with org.jline:jansi
### New feature, improvement proposal

From #372

> This library is no longer maintained. It has been merged into org.jline:jansi.
Please update your dependencies to use org.jline:jansi instead.

HarshMehta112:
Fixes #373

### Changes
- Updated `org.jline:jansi` to version `3.25.1` (last version supporting Java 8)
- Ensured compatibility with existing usage in `maven-shared-utils`

### Notes
- Versions of `jansi` newer than `3.25.1` require Java 11+, so they cannot be used while maintaining Java 8 support.
- This is a dependency-only change with no functional impact expected.

### Summary
This update preserves Java 8 compatibility while keeping the dependency as up to date as possible within supported constraints.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,10 +74,9 @@
     </dependency>
 
     <dependency>
-      <groupId>org.fusesource.jansi</groupId>
+      <groupId>org.jline</groupId>
       <artifactId>jansi</artifactId>
-      <version>2.4.3</version>
-      <optional>true</optional>
+      <version>3.25.1</version>
     </dependency>
     <dependency>
       <groupId>org.jspecify</groupId>
--- a/src/main/java/org/apache/maven/shared/utils/logging/AnsiMessageBuilder.java
+++ b/src/main/java/org/apache/maven/shared/utils/logging/AnsiMessageBuilder.java
@@ -18,7 +18,7 @@
  */
 package org.apache.maven.shared.utils.logging;
 
-import org.fusesource.jansi.Ansi;
+import org.jline.jansi.Ansi;
 
 /**
  * Message builder implementation that supports ANSI colors through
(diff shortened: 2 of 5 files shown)


slawekjaranowski (changes requested):
It is no so easy, this class is used by Maven until 3.9.x and many plugins ....

Maven also exports such packages 

In Maven 3.10.x we will MessageBuilderFactory class.

So I will not touch it, we can make it deprecated.

slachiewicz:
Closing this in favour of deprecating the package instead, per the review here.

The blocker is that `org.apache.maven.shared.utils.logging` is exported by Maven up to 3.9.x and many plugins compile against it, so changing the ANSI provider underneath it is not a safe change to make in this library. Maven 3.10.0 introduces `org.apache.maven.message.MessageBuilderFactory` (and Maven 4 `org.apache.maven.api.services.MessageBuilderFactory`) as the replacement, so the path forward is to point users there rather than re-plumb this code.

That deprecation is now up as [PR #423](https://github.com/apache/maven-shared-utils/pull/423), which also closes [#373](https://github.com/apache/maven-shared-utils/issues/373).

Thanks for the patch, and sorry it did not land -- the constraint only surfaced in review. The Java 8 analysis you did on `org.jline:jansi` 3.25.1 is still useful and is worth keeping on [#373](https://github.com/apache/maven-shared-utils/issues/373) for whoever revisits the provider question in a Maven 4-only line.

Your branch `fix/#373` is on your fork, so please delete it there when convenient -- committers cannot remove it from here.

PR #391, opened 2026-07-01
XMLEncode: encode illegal XML control characters

### LINKED ISSUE #390: XMLEncode: illegal XML control characters not encoded
`XMLEncode.xmlEncodeTextAsPCDATA()` passes characters in the range U+0000-U+001F (excluding U+0009 TAB, U+000A LF, U+000D CR) through unencoded in the `default` branch of its switch statement. These characters are illegal in XML 1.0 and will cause XML parsers to reject the output.

The `default` case at line 112 just does `n.append(c)` — it should instead encode these characters as `&#xHH;` numeric character references. The explicitly handled cases (`&`, `<`, `>`, `"`, `'`, `\r`, `\n`) cover the legal control chars, but `\0`, `\1`-`\b`, `\v`, `\f`, `\u000E`-`\u001F` all slip through.

This affects both attribute values (via `PrettyPrintXMLWriter.addAttribute()`) and text content (via `PrettyPrintXMLWriter.writeText()`).

Fix: add a check in the `default` case to encode illegal XML control characters.

elharo:
`XMLEncode.xmlEncodeTextAsPCDATA()` passes characters in the range U+0000–U+001F (excluding TAB, LF, CR) through unencoded in the `default` branch of its switch statement. These characters are illegal in XML 1.0 and cause XML parsers to reject the output.

Additionally, `needsEncoding()` only checked for `&` and `<`, so text containing only control characters was written directly without reaching the encoding method at all.

**Fix:**
- Throw IOException for completely illegal characters that cannot appear in XML 
- In `needsEncoding()`, added a check for illegal control chars so they're routed through the encoding path

Fixes https://github.com/apache/maven-shared-utils/issues/390

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/xml/XMLEncode.java
+++ b/src/main/java/org/apache/maven/shared/utils/xml/XMLEncode.java
@@ -109,7 +109,13 @@ final class XMLEncode {
                     break;
 
                 default:
-                    n.append(c);
+                    if (c < 0x20 && c != 0x09 && c != 0x0A && c != 0x0D) {
+                        n.append("&#x");
+                        n.append(Integer.toHexString(c));
+                        n.append(';');
+                    } else {
+                        n.append(c);
+                    }
                     break;
             }
         }
@@ -143,6 +149,9 @@ final class XMLEncode {
         for (int i = 0; i < text.length(); i++) {
             char c = text.charAt(i);
             if (c == '&' || c == '<') {
+                return true;
+            }
+            if (c < 0x20 && c != 0x09 && c != 0x0A && c != 0x0D) {
                 return true;
             }
         }
--- a/src/main/java/org/apache/maven/shared/utils/xml/XMLEncode.java
+++ b/src/main/java/org/apache/maven/shared/utils/xml/XMLEncode.java
@@ -109,6 +109,12 @@ final class XMLEncode {
                     break;
 
                 default:
+                    // C0 control characters (except tab, LF, CR) are encoded
+                    // as numeric character references. This produces valid
+                    // XML 1.1 but is not valid XML 1.0 (which forbids these
+                    // characters in any form). Callers that require strict
+                    // XML 1.0 compliance should strip these characters before
+                    // encoding.
                     if (c < 0x20 && c != 0x09 && c != 0x0A && c != 0x0D) {
                         n.append("&#x");
                         n.append(Integer.toHexString(c));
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/shared/utils/xml/XMLEncode.java
```diff
@@ -109,7 +109,13 @@ static void xmlEncodeTextAsPCDATA(String text, boolean forAttribute, char quoteC
                     break;
 
                 default:
-                    n.append(c);
+                    if (c < 0x20 && c != 0x09 && c != 0x0A && c != 0x0D) {
```
elharo, line 114: This only works in XML 1.1. Perhaps we should thrown an exception in this case.

slachiewicz:
Merged current `master` into this branch to clear the conflict.

The conflict looked worse than it was: [#382](https://github.com/apache/maven-shared-utils/pull/382) and this branch each appended a test method at the same point in `PrettyPrintXmlWriterTest`, so git could not tell them apart. Both are kept as-is -- nothing semantic was resolved and no assertion was changed. `master` also now carries the [#422](https://github.com/apache/maven-shared-utils/pull/422) fix, so the branch builds green again.

`XMLEncode` itself is untouched.

Verified: `mvn -B verify` on JDK 17 -> Tests run: 789, Failures: 0, Errors: 0. Spotless clean.

This is still a draft. From here it looks ready to mark for review unless you had more planned for the C0 control character handling.

elharo (commented):
This PR is a work in progress. I simply haven't decided what should happen in this case. A good approach that maintains API compatibility is not obvious.

elharo:
OK, I think we can just throw an IOException here and then never emit unencodable characters

src/main/java/org/apache/maven/shared/utils/xml/XMLEncode.java
elharo: or maybe an IllegalArgumnetException?

slachiewicz (commented):
Not mergeable yet: the CDATA path still passes the characters through. `xmlEncodeText` routes any text longer than 12 characters without `]]>` to `xmlEncodeTextAsCDATABlock`, which never reaches the new check, so the writer emits the control character inside a CDATA block with no exception. CDATA does not make U+0001 legal in XML 1.0.

Reproduction on this branch: `writeText("hello" + (char) 1 + "world, this is longer than twelve chars")` → `<div><![CDATA[hello{U+0001}world, ...]]>` and no exception; the same text cut to three characters throws as intended. The check needs to sit before the CDATA branch, or in `xmlEncodeText` itself.

Two smaller points. The `needsEncoding` javadoc still says the characters are encoded as numeric references producing XML 1.1 output, which the last commit replaced with a throw. On the open exception-type thread, `IllegalArgumentException` fits better: the input is wrong rather than the stream, and callers of `XMLWriter` already handle `IOException` for other reasons, so a bad input would go unnoticed.

PR #393, opened 2026-07-01
Xpp3DomBuilder.createXmlReader: remove thread-unsafe system property manipulation

### LINKED ISSUE #392: Xpp3DomBuilder.createXmlReader() system property race condition
`Xpp3DomBuilder.createXmlReader()` (lines 119-129) clears and restores the global `org.xml.sax.driver` system property without synchronization. In a multi-threaded environment (like Maven parallel builds), one thread's clear/restore sequence can interfere with another thread's read or write of the same property. The code even acknowledges this with the comment: "There's a 'slight' problem with this an parallel maven: It does not work ;)"

The method already tries to directly instantiate `com.sun.org.apache.xerces.internal.parsers.SAXParser` first (line 114), which succeeds on all Oracle/OpenJDK JVMs. The system property manipulation fallback is both unnecessary and harmful.

Fix: remove the system property manipulation. If the direct instantiation fails, just call `XMLReaderFactory.createXMLReader()` without modifying any system properties.

elharo:
`Xpp3DomBuilder.createXmlReader()` clears and restores the global `org.xml.sax.driver` system property without synchronization (lines 119-129). In a multi-threaded environment like Maven parallel builds, one thread's clear/restore sequence interferes with other threads' reads of the same property. The code's own comment acknowledges: "There's a 'slight' problem with this an parallel maven: It does not work ;)"

The method already tries to directly instantiate `com.sun.org.apache.xerces.internal.parsers.SAXParser` first — which succeeds on all Oracle/OpenJDK JVMs. The system property manipulation fallback is both unnecessary and harmful.

**Fix:** Removed the system property manipulation entirely. If the direct instantiation fails, falls through to `XMLReaderFactory.createXMLReader()` without modifying any global state.

Fixes https://github.com/apache/maven-shared-utils/issues/392


slachiewicz:
Pushed two commits to this branch: merged current `master` (which now carries the [#422](https://github.com/apache/maven-shared-utils/pull/422) fix, so the branch builds green again), and addressed the review point on the new test.

The test now restores `org.xml.sax.driver` in a `finally` block. `build()` must not touch that property, which is exactly what the test asserts, but if that ever regresses the old form would leave the mutated value behind and take unrelated tests down with it. This keeps the blast radius to this test, matching how `OsTest` handles system properties here.

No change to `Xpp3DomBuilder` itself.

Verified: `mvn -B verify` on JDK 17 -> Tests run: 788, Failures: 0, Errors: 0. Spotless clean.

slachiewicz (approved):
The race is real and removing the clear/restore is the right fix: a thread reading `org.xml.sax.driver` during another thread's window saw it cleared, and no restore ordering can make that safe.

Two corrections to the description, for the record. On JDK 17 and 25 the direct `com.sun.org.apache.xerces.internal.parsers.SAXParser` instantiation fails with `IllegalAccessException`, because `java.xml` does not export that package, so `XMLReaderFactory.createXMLReader()` is the path actually taken there rather than a rarely used fallback. The fix still holds: the factory returns the same Xerces parser when the property is unset. The one behaviour change is that a user-set `org.xml.sax.driver` is now honoured, including a broken value, where the old code silently ignored it. That is what the property is documented to do.

The new test also passes on the old code, since with the property unset the old clear and restore were no-ops. It guards against a future regression rather than demonstrating this one.

Squash-merging, since the branch carries a master merge and two fixups.

Verified: probe on JDK 17 and 25 → direct instantiation throws `IllegalAccessException`, factory returns `com.sun.org.apache.xerces.internal.parsers.SAXParser`; CI green on all 12 jobs.

PR #397, opened 2026-07-01
Xpp3DomWriter: add explicit null check for dom parameter

### LINKED ISSUE #396: Xpp3DomWriter.write(null) throws confusing NPE
All public `write` methods in `Xpp3DomWriter` accept an `Xpp3Dom dom` parameter documented as `@param dom Xpp3Dom` but without null checks. Passing `null` for `dom` causes a `NullPointerException` at line 65 (`dom.getName()`).

Fix: add null check (with clear message) at the shared `write(XMLWriter, Xpp3Dom, boolean)` method.

elharo:
All public `write` methods accept an `Xpp3Dom dom` parameter without null checks. Passing `null` throws a `NullPointerException` deep inside the delegation chain (at `dom.getName()` in the `write(XMLWriter, Xpp3Dom, boolean)` method), producing a confusing stack trace for callers.

**Fix:** Added `Objects.requireNonNull(dom, "dom must not be null")` at the entry to the shared `write(XMLWriter, Xpp3Dom, boolean)` method, so callers get a clear message immediately.

Fixes https://github.com/apache/maven-shared-utils/issues/396

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomWriter.java
+++ b/src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomWriter.java
@@ -21,6 +21,7 @@ package org.apache.maven.shared.utils.xml;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.Writer;
+import java.util.Objects;
 
 /**
  * @author Brett Porter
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/shared/utils/xml/Xpp3DomWriter.java
```diff
@@ -62,6 +63,7 @@ public static void write(XMLWriter xmlWriter, Xpp3Dom dom) throws IOException {
      * @throws IOException if writing fails
      */
     public static void write(XMLWriter xmlWriter, Xpp3Dom dom, boolean escape) throws IOException {
+        Objects.requireNonNull(dom, "dom must not be null");
```
desruisseaux, line 66: Is it really solving the problem described in the issue? The next line is `dom.getName()`, therefore a `NullPointerException` would happen there anyway. The stack trace produced by `Objects` will be even deeper (a little bit) than the stack trace produced by the `NullPointerException` on the next line.

elharo: makese sense. thanks.

PR #399, opened 2026-07-01
CommandLineUtils: ensure all pumpers are waited on in try-finally

### LINKED ISSUE #398: CommandLineUtils.call() does not ensure all pumpers are waited on
In `CommandLineUtils`, the `call()` method calls `waitUntilDone()` on `inputFeeder`, `outputPumper`, and `errorPumper` sequentially (lines 298-303). If the first call throws (e.g., `InterruptedException`), the remaining pumpers are never waited on, leaving their threads potentially running.

The commented-out code above (lines 277-297) shows the original author intended a try-finally chain, but it was disabled.

Fix: wrap the calls in a try-finally to ensure all pumpers are waited on even when an earlier one throws.

elharo:
In `CommandLineUtils.call()`, `waitUntilDone()` was called on `inputFeeder`, `outputPumper`, and `errorPumper` sequentially without try-finally (lines 298-303 of the original). If an earlier call threw (e.g., `InterruptedException`), the remaining pumpers were never waited on, leaving their threads potentially running.

The commented-out code just above (lines 277-297) shows the original author intended a try-finally chain but never enabled it.

**Fix:** Replaced the sequential calls and the dead commented block with a proper try-finally chain ensuring all three pumpers are waited on regardless of exceptions from earlier ones.

Fixes https://github.com/apache/maven-shared-utils/issues/398

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
@@ -274,33 +274,17 @@ public abstract class CommandLineUtils {
 
                     int returnValue = p.waitFor();
 
-                    // TODO Find out if waitUntilDone needs to be called using a try-finally construct. The method may
-                    // throw an
-                    //      InterruptedException so that calls to waitUntilDone may be skipped.
-                    //                    try
-                    //                    {
-                    //                        if ( inputFeeder != null )
-                    //                        {
-                    //                            inputFeeder.waitUntilDone();
-                    //                        }
-                    //                    }
-                    //                    finally
-                    //                    {
-                    //                        try
-                    //                        {
-                    //                            outputPumper.waitUntilDone();
-                    //                        }
-                    //                        finally
-                    //                        {
-                    //                            errorPumper.waitUntilDone();
-                    //                        }
-                    //                    }
-                    if (inputFeeder != null) {
-                        inputFeeder.waitUntilDone();
+                    try {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
```diff
@@ -274,34 +274,18 @@ public Integer call() throws CommandLineException {
...
+                    try {
+                        if (inputFeeder != null) {
+                            inputFeeder.waitUntilDone();
+                        }
+                    } finally {
+                        try {
+                            outputPumper.waitUntilDone();
+                        } finally {
+                            errorPumper.waitUntilDone();
+                        }
```
desruisseaux, line 286: It works, but if more than one exception is thrown, we see only the last one. It would be possible to keep all of them with `catch (Throwable e)` and `Throwable.addSuppressed(e)` calls, but that would make the code more complex (I'm not sure it would be worth). Alternatively, this exception handling could be made easier if the pumpers implement `AutoCloseable`.

I'm not really suggesting a change. I'm not against deciding that it is not worth to keep all exceptions. Just submitting for your consideration and letting you decide.

desruisseaux (approved):

src/main/java/org/apache/maven/shared/utils/cli/CommandLineUtils.java
elharo: Makes sense, but I think I;m going to leave that for a future PR

PR #403, opened 2026-07-01
PathTool.getRelativeFilePath: return empty string instead of null for cross-drive paths

### LINKED ISSUE #402: PathTool.getRelativeFilePath returns null instead of empty string for cross-drive paths
`PathTool.getRelativeFilePath()` returns `null` for Windows cross-drive scenarios (lines 167 and 174) instead of `""` (empty string). The method's Javadoc shows `""` for null/empty inputs, and callers may not expect null.

Fix: change the two `return null` to `return ""`.

elharo:
Fixes https://github.com/apache/maven-shared-utils/issues/402

### DIFF
--- a/src/test/java/org/apache/maven/shared/utils/PathToolTest.java
+++ b/src/test/java/org/apache/maven/shared/utils/PathToolTest.java
@@ -57,10 +57,12 @@ public class PathToolTest {
         assertEquals("../../bin", PathTool.getRelativeFilePath("/usr/local/", "/bin"));
 
         assertEquals("../usr/local/", PathTool.getRelativeFilePath("/bin", "/usr/local/"));
+
+        assertEquals("", PathTool.getRelativeFilePath("c:/usr/local", "d:/java/bin"));
     }
 
     @Test
-    // Keep in sync with testGetRelativeFilePathNonWindows()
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/maven/shared/utils/PathToolTest.java
```diff
@@ -57,10 +57,12 @@ public void testGetRelativeFilePathNonWindows() {
         assertEquals("../../bin", PathTool.getRelativeFilePath("/usr/local/", "/bin"));
 
         assertEquals("../usr/local/", PathTool.getRelativeFilePath("/bin", "/usr/local/"));
+
+        assertEquals("", PathTool.getRelativeFilePath("c:/usr/local", "d:/java/bin"));
     }
 
     @Test
-    // Keep in sync with testGetRelativeFilePathNonWindows()
+    // Keep in sync with testGetRelativeFilePathWindows()
```
elharo: I think this comment was accurate and should not be changed

desruisseaux (approved):
I'm not familiar with the context in which this method was used. But it is a deprecated method anyway, so hopefully the method is not widely used and therefore the risk is low.

PR #407, opened 2026-07-01
Os.getOsFamily: return empty string instead of null

### LINKED ISSUE #406: Os.getOsFamily() returns null when no family matches
`Os.getOsFamily()` returns `null` (line 368) when no known OS family matches the current `os.name`. The `OS_FAMILY` constant (line 75) is then `null`, which can cause NPE in callers that expect a non-null string.

Fix: return `""` instead of `null` when no family matches.

elharo:
Fixes https://github.com/apache/maven-shared-utils/issues/406

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/Os.java
+++ b/src/main/java/org/apache/maven/shared/utils/Os.java
@@ -365,7 +365,7 @@ public class Os {
                 return fam;
             }
         }
-        return null;
+        return "";
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/shared/utils/Os.java
```diff
@@ -365,7 +365,7 @@ private static String getOsFamily() {
                 return fam;
             }
         }
-        return null;
```
sparsick, line 368: I didn't find a contract, that ensures this piece of code should not be null. I'm afraid, that could have side effects in code that uses it. 

For example: 

```java
if (Os.OS_FAMILY == null) {
}
```

This snippet would have another meaning after our change. 

On the other side, the class `Os` is deprecated, so maybe it is okay to take this risk.

elharo: Makes sense. probably better not to take it then.

PR #412, opened 2026-07-01
Close process streams after exit to work around JDK-4311711 hang

### LINKED ISSUE #270: [MSHARED-688] [WORKAROUND] executeCommandLineAsCallable hangs reading open std/out on exited Process as JVM bug
**[Tibor Digana](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=tibor17)** opened **[MSHARED-688](https://issues.apache.org/jira/browse/MSHARED-688?redirect=false)** and commented

Need additional parameter to force closing out and error stream after Process exit. The parameter in the old method `waitForAllPumpers( inputFeeder, outputPumper, errorPumper )` and the enclosing method `executeCommandLineAsCallable()`.
Java does not close `FileInputStream`. The native stream is mapped to file system.
Reproduced with `maven-shared-utils:0.9`
This is the thread dump:

Main Thread:
Stack trace:
java.lang.Object.wait(Native Method)
java.lang.Object.wait(Object.java:485)
org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.AbstractStreamHandler.waitUntilDone(AbstractStreamHandler.java:42)
org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.StreamPumper.waitUntilDone(StreamPumper.java:42)
org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.CommandLineUtils.waitForAllPumpers(CommandLineUtils.java:339)
org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.CommandLineUtils.access$200(CommandLineUtils.java:41)
org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.CommandLineUtils$1.call(CommandLineUtils.java:285)
org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:616)
org.apache.maven.plugin.surefire.booterclient.ForkStarter.fork(ForkStarter.java:533)
org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:278)
org.apache.maven.plugin.surefire.booterclient.ForkStarter.run(ForkStarter.java:244)

Name: Thread-3
State: RUNNABLE
Total blocked: 0  Total waited: 0

Stack trace:
java.io.FileInputStream.readBytes(Native Method)
java.io.FileInputStream.read(FileInputStream.java:220)
java.io.BufferedInputStream.read1(BufferedInputStream.java:256)
java.io.BufferedInputStream.read(BufferedInputStream.java:317)
- locked java.io.BufferedInputStream@3c640f1a
  sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:264)
  sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:306)
  sun.nio.cs.StreamDecoder.read(StreamDecoder.java:158)
- locked java.io.InputStreamReader@4d1b92ef
  java.io.InputStreamReader.read(InputStreamReader.java:167)
  java.io.BufferedReader.fill(BufferedReader.java:136)
  java.io.BufferedReader.readLine(BufferedReader.java:299)
- locked java.io.InputStreamReader@4d1b92ef
  java.io.BufferedReader.readLine(BufferedReader.java:362)
  org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.StreamPumper.run(StreamPumper.java:76)

Name: Thread-4
State: RUNNABLE
Total blocked: 0  Total waited: 0

Stack trace:
java.io.FileInputStream.readBytes(Native Method)
java.io.FileInputStream.read(FileInputStream.java:220)
sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:264)
sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:306)
sun.nio.cs.StreamDecoder.read(StreamDecoder.java:158)
- locked java.io.InputStreamReader@2b87514a
  java.io.InputStreamReader.read(InputStreamReader.java:167)
  java.io.BufferedReader.fill(BufferedReader.java:136)
  java.io.BufferedReader.readLine(BufferedReader.java:299)
- locked java.io.InputStreamReader@2b87514a
  java.io.BufferedReader.readLine(BufferedReader.java:362)
  org.apache.maven.surefire.shade.org.apache.maven.shared.utils.cli.StreamPumper.run(StreamPumper.java:76)

---
No further details from [MSHARED-688](https://issues.apache.org/jira/browse/MSHARED-688?redirect=false)

elharo:
After `Process.waitFor()` returns, close the process's stdout, stderr, and stdin streams. This works around [JDK-4311711](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4311711) where `Process.getInputStream().read()` can hang indefinitely even after the process has terminated.

Without this fix, `CommandLineUtils.executeCommandLineAsCallable()` can hang forever on some JVMs because the `StreamPumper` threads are blocked in `BufferedReader.readLine()` on the process's output streams, which never reach EOF.

Fixes https://github.com/apache/maven-shared-utils/issues/270

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -24,4 +24,4 @@ on:
 jobs:
   build:
     name: Verify
-    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v4
+    uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v5
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -24,4 +24,4 @@ on:
 jobs:
   pr-automation:
     name: PR Automation
-    uses: apache/maven-gh-actions-shared/.github/workflows/pr-automation.yml@v4
+    uses: apache/maven-gh-actions-shared/.github/workflows/pr-automation.yml@v5
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -24,4 +24,4 @@ on:
 
 jobs:
    update_release_draft:
-      uses: apache/maven-gh-actions-shared/.github/workflows/release-drafter.yml@v4
+      uses: apache/maven-gh-actions-shared/.github/workflows/release-drafter.yml@v5
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -25,4 +25,4 @@ on:
 
 jobs:
   stale:
-    uses: 'apache/maven-gh-actions-shared/.github/workflows/stale.yml@v4'
+    uses: 'apache/maven-gh-actions-shared/.github/workflows/stale.yml@v5'
--- a/src/main/java/org/apache/maven/shared/utils/StringUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/StringUtils.java
@@ -963,6 +963,9 @@ public class StringUtils {
      */
     @NonNull
     public static String chopNewline(@NonNull String str) {
+        if (str.isEmpty()) {
+            return "";
+        }
         int lastIdx = str.length() - 1;
         char last = str.charAt(lastIdx);
         if (last == '\n') {
@@ -1108,11 +1111,13 @@ public class StringUtils {
      * @param size  size to pad to
      * @param delim string to pad with
      * @return right padded String
-     * @throws ArithmeticException  if delim is the empty String
      * @throws NullPointerException if str or delim is <code>null</code>
      */
     @NonNull
     public static String rightPad(@NonNull String str, int size, @NonNull String delim) {
+        if (delim.isEmpty()) {
+            return str;
+        }
         size = (size - str.length()) / delim.length();
         if (size > 0) {
             str += repeat(delim, size);
@@ -1142,11 +1147,13 @@ public class StringUtils {
      * @param size  size to pad to
      * @param delim string to pad with
      * @return left padded String
-     * @throws ArithmeticException  if delim is the empty string
      * @throws NullPointerException if str or delim is null
      */
     @NonNull
     public static String leftPad(@NonNull String str, int size, @NonNull String delim) {
+        if (delim.isEmpty()) {
+            return str;
+        }
         size = (size - str.length()) / delim.length();
         if (size > 0) {
             str = repeat(delim, size) + str;
--- a/src/main/java/org/apache/maven/shared/utils/cli/AbstractStreamHandler.java
+++ b/src/main/java/org/apache/maven/shared/utils/cli/AbstractStreamHandler.java
@@ -45,7 +45,7 @@ class AbstractStreamHandler extends Thread {
      * @return {@code true} if the handler finished within the timeout, {@code false} otherwise
      * @throws InterruptedException if the current thread is interrupted while waiting
      */
-    public synchronized boolean waitUntilDone(long timeoutInMillis) throws InterruptedException {
+    synchronized boolean waitUntilDone(long timeoutInMillis) throws InterruptedException {
         if (timeoutInMillis <= 0) {
             waitUntilDone();
             return true;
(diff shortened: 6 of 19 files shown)


ascheman (changes requested):
Thanks for tackling this @elharo — #270 is a real hang and the instinct (force EOF by closing the stream) is right for the *stuck* case. But I ran an empirical check against this branch vs `master`, and the unconditional close introduces a spurious-failure regression on the **normal** path.

**Setup:** `seq 1 5000` (exits 0, 5000 lines) via `executeCommandLine`, 60 runs, counting non-zero exits, truncated output, and thrown `CommandLineException`. Same probe on `master` and on this branch, JDK 17 and JDK 25.0.3:

| branch | JDK 17 | JDK 25.0.3 |
|---|---|---|
| `master` | 0/60 exceptions | 0/60 |
| this PR | 6–7/60 (~10%) | 7/60 (~12%) |

All failures are `CommandLineException: Failure processing stdout` on a command that exited 0 with complete output (no truncation observed, so it's not data loss — it's a false failure).

**Why:** after `waitFor()` returns, the pumper threads are often still draining buffered output. Closing `getInputStream()`/`getErrorStream()` out from under an in-flight `readLine()` makes it throw `IOException`, which `StreamPumper.run()` stores in `exception`, and `call()` then rethrows as `CommandLineException`. A separate probe confirms that when the pumper is idle-blocked on an *empty* stream (the actual JDK-4311711 hang), the close yields a clean EOF with no exception — so the close only misfires when there's in-flight data.

**Suggested direction:** make the close a *fallback*, not the default — `waitUntilDone(timeout)` first (pumpers reach EOF on their own in the common case, as `master` shows), and only force-close if they're still stuck past the timeout. Additionally, mark the pumper as intentionally-closed (there's already a `disable()` hook) so an `IOException` arising from *our* close is treated as EOF rather than recorded as a stdout/stderr failure. That keeps the hang fix while removing the false failures.

Happy to contribute the probe as a proper regression test if useful — it reliably reproduces the ~10% false-failure on this branch and stays green on `master`.

elharo:
If the probe can be contributed as an integration or unit test that would be very useful. Thanks. Does it need to be part of this PR or could it be sent against master?

elharo:
Thanks for the detailed review — the probe reproducing the ~10% spurious 'Failure processing stdout' on a cleanly-exiting process was very helpful. I've merged master into this branch and reworked the close into a fallback as suggested:

* The process streams are no longer closed unconditionally. After `waitFor()` the pumpers are first given a grace period (`STREAM_EOF_GRACE_PERIOD_MS = 5000`) to drain the remaining output and reach EOF on their own (the common case, as on master). Only if they are still stuck after the grace period do we force EOF by closing the streams.
* Before forcing the close, both pumpers are marked with the existing `disable()` hook, and `StreamPumper.run()` now treats an `IOException` as EOF when disabled, so the close no longer gets recorded as a stdout/stderr failure.
* Added a timeout-aware `AbstractStreamHandler.waitUntilDone(long)` for the grace-period wait (plus unit tests for it).
* Added a regression test (`executeCommandLineWithLargeStdoutCompletesWithoutFailure`) based on your probe: it runs `seq` via `executeCommandLine` repeatedly and asserts exit code 0, complete output, and no exception. I verified it reproduces the spurious `CommandLineException` against the previous unconditional-close code and stays green with the fallback. It is Unix-only (`seq`) and skipped elsewhere.

The PR is now up to date with master and green on the matrix. Happy to adjust the grace-period value or anything else.

ascheman (approved):
Thanks @elharo — this is exactly it. The grace-period `waitUntilDone` lets the pumpers reach EOF on their own in the normal case, and disabling them before the fallback close (with `run()` now skipping `exception` when disabled) means the close-induced `IOException` is treated as EOF rather than a spurious `CommandLineException`. Reran my earlier probe on this head (`seq 1 5000` × 60): **0/60 spurious failures on both JDK 17 and 25** — matching `master`, vs ~10–12% on the previous unconditional-close version. Regression gone. LGTM.

PR #417, opened 2026-07-23
Fix documentation issues in site pages

elharo:
Various documentation fixes:

- Upgrade plexus-utils link to HTTPS
- Reword "forked from" → "originated in" (more precise)
- Clarify ambiguous "the original" phrasing
- Remove redundant parenthetical "(different)"
- Use proper project reference "Commons versions"
- Use project display name "Maven Shared Utils"
- Upgrade XSD URL from HTTP to HTTPS in site.xml

### DIFF
--- a/src/site/apt/index.apt.vm
+++ b/src/site/apt/index.apt.vm
@@ -40,7 +40,7 @@ ${project.name}
 
 Why?
 
-   plexus-utils consisted mostly of code that originated in various Apache projects. 
+   plexus-utils consisted mostly of code that originated in various Apache projects.
    maven-shared-utils is based on those original Apache-source implementations.
 
 Why not commons?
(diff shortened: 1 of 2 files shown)


ascheman (approved):
Thanks Elliott — nice cleanup. I checked each change:

- **HTTPS upgrades** (plexus-utils link in `index.apt.vm`, XSD `schemaLocation` in `site.xml`) — correct. Good catch leaving the `xmlns=` decoration namespace on `http://`; that's an identifier, not a URL, so it must stay as-is.
- **Wording** — "originated in" over "forked from", dropping the redundant "(different)", and "Commons versions" all read more precisely.
- **Naming** — standardizing the prose on "Maven Shared Utils" (the `${project.name}` display form) is the right call; both remaining prose mentions are now consistent.

Copilot's consistency note is already satisfied by the final state of the file — no lowercase `maven-shared-utils` left in prose. CI is green across all 11 jobs. LGTM. 👍

PR #421, opened 2026-08-09
Restore the common wording on the download page

slachiewicz:
The wording on this page had drifted from the text the rest of the Maven estate uses. This restores the shared wording so the page is identical to its counterparts apart from the distribution area and the artifact name.

Format is unchanged — this stays xdoc.

### This is not purely cosmetic

Diffing the rendered HTML before and after shows the reworded version had **silently dropped three hyperlinks**, leaving the text as plain prose:

- `https://www.apache.org/info/verification.html` (how to verify a download)
- `https://downloads.apache.org/maven/KEYS` (the signing keys)
- `https://archive.apache.org/dist/maven/…/` (the archive site)

Restoring the common wording restores all three.

Headings, generated anchors (`#Files`, `#Previous_Versions`) and the table are unchanged, so no URL and no in-page anchor moves.

Verified by local site build and HTML diff only — not claiming CI is green; draft until it is.

---

**Licence header:** left byte-for-byte as it was. My first pass spliced in a canonical header and silently downgraded an `https://` LICENSE-2.0 URL to `http://` in a sibling repo; the generator now preserves each file's existing preamble and never rewrites it.

**Format:** this stays xdoc. Converting the page to Markdown was evaluated separately and shown to round-trip losslessly, and was then deliberately declined — noting that here so the question does not get reopened from scratch later.

### DIFF
--- a/src/site/xdoc/download.xml.vm
+++ b/src/site/xdoc/download.xml.vm
@@ -29,14 +29,14 @@ under the License.
 
       <p><strong>${project.name} ${project.version}</strong> is distributed in source format.</p>
 
-      <p>Use a source archive if you want to build ${project.name} yourself.</p>
+      <p>Use a source archive if you intend to build <strong>${project.name}</strong> yourself.</p>
 
-      <p>Otherwise, use the ready-made binary artifacts from the central repository.</p>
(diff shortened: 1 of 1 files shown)


elharo (changes requested):
The change you're trying to revert was a deliberate improvement.

PR #424, opened 2026-09-10
PropertyUtils: stop closing the caller's InputStream

### LINKED ISSUE #274: [MSHARED-895] PropertiesUtil closes streams it doesn't create
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MSHARED-895](https://issues.apache.org/jira/browse/MSHARED-895?redirect=false)** and commented

specifically the methods loadOptionalProperties(InputStream inputStream ) and loadProperties(InputStream inputStream )

 

see discussion on https://github.com/apache/maven-shared-utils/pull/41

---

1 votes, 2 watchers

slachiewicz:
`PropertyUtils.loadProperties(InputStream)` and `loadOptionalProperties(InputStream)` close the stream the caller handed them, via a try-with-resources that re-binds the parameter. A method that did not open a stream should not close it -- the caller may still need it, and closing it is invisible from the signature.

Fixes [#274](https://github.com/apache/maven-shared-utils/issues/274).

The commit is [Elliotte Rusty Harold](https://github.com/elharo)'s, cherry-picked from the `fix/propertyutils-stream-close` branch he pushed here on 2026-07-01 and never opened as a PR. Authorship is preserved. The only thing added on top is `spotless:apply` output, since that branch predates a formatting change on `master`. Opening it because the fix is finished and 3.5.0 is close; happy to hand it back if he would rather carry it himself.

Note for reviewers: this is a behaviour change. Callers that relied on the accidental close now have to close their own stream. That is the point of the issue, but it is worth a moment's thought before merging.

Verified: `mvn -B verify` on JDK 17 -> Tests run: 789, Failures: 0, Errors: 0. Spotless clean.

### DIFF
--- a/src/main/java/org/apache/maven/shared/utils/PropertyUtils.java
+++ b/src/main/java/org/apache/maven/shared/utils/PropertyUtils.java
@@ -90,11 +90,7 @@ public class PropertyUtils {
         try {
             Properties result = new Properties();
             if (is != null) {
-                try (InputStream in = is) {
-                    result.load(in);
-                } catch (IOException e) {
-                    // ignore
-                }
+                result.load(is);
             }
             return result;
         } catch (Exception e) {
@@ -168,9 +164,8 @@ public class PropertyUtils {
         Properties properties = new Properties();
 
         if (inputStream != null) {
-            try (InputStream in = inputStream) // reassign inputStream to autoclose
-            {
-                properties.load(in);
+            try {
+                properties.load(inputStream);
             } catch (IllegalArgumentException | IOException ex) {
                 // ignore and return empty properties
             }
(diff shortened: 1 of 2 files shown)


elharo (commented):
I vaguely recall this. I think there was some reason I didn't open a PR, likely I wasn't convinced this was correct or wouldn't break anything.

src/main/java/org/apache/maven/shared/utils/PropertyUtils.java
```diff
@@ -168,9 +164,8 @@ public static Properties loadOptionalProperties(final @Nullable InputStream inpu
         Properties properties = new Properties();
 
         if (inputStream != null) {
-            try (InputStream in = inputStream) // reassign inputStream to autoclose
-            {
-                properties.load(in);
+            try {
+                properties.load(inputStream);
             } catch (IllegalArgumentException | IOException ex) {
                 // ignore and return empty properties
```
elharo, line 170: The real problem might be here. Failure to read properties due to I/O errors should be an exception. We shouldn't just return empty properties

PR #429, opened 2026-09-10
Drop the jspecify nullness annotations

slachiewicz:
Removes the `org.jspecify:jspecify` dependency and every `@NonNull` / `@Nullable` annotation it supplied, in place of the provided-scope change this PR first proposed. The review on that version asked whether the annotations earn their place at all; they are not checked by any tool in this build, so dropping them is the smaller change to the consumer graph than adjusting the scope.

The diff is large only because spotless re-joins signatures that fit on one line once the annotations are gone. No method body changes.

What consumers see: the runtime annotation metadata disappears from public signatures. Nothing in this library reflects on it, Maven core has no reference to `org.jspecify` at all, and NullAway or similar checkers were never wired into this build. I have not audited the other consumers; one that referenced `org.jspecify` transitively through this artifact would need to declare it itself.

After the change `dependency:list -DincludeScope=runtime` yields `slf4j-api`, `jansi` (optional) and `commons-io`.

Verified: `JAVA_HOME=$(/usr/libexec/java_home -v 17) mvn -B verify` -> Tests run: 788, Failures: 0, Errors: 0 (JDK 17); spotless clean; `dependency:list -DincludeScope=runtime` shows no jspecify

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,11 +79,6 @@
       <version>2.4.3</version>
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>org.jspecify</groupId>
-      <artifactId>jspecify</artifactId>
-      <version>1.0.1</version>
-    </dependency>
 
     <dependency>
       <groupId>commons-io</groupId>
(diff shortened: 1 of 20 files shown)


elharo (approved):
This makes me wonder if we could just drop this. The 18 different variants of Null/NonNull/Nullable etc. feel like an experiment that has failed. I've never seem them improve robustness in any meaningful way. They're applied haphazardly and never checked.

Issue #286, opened 2019-08-29
[MSHARED-832] Cannot invoke maven from path with parentheses on Windows


jira-importer:
**[Anton Krylov](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=aek)** opened **[MSHARED-832](https://issues.apache.org/jira/browse/MSHARED-832?redirect=false)** and commented

When `M2_HOME` is set to a path with parentheses (in this case `C:\work\lol(1)\maven)`, the following command is generated on windows:

```
cmd.exe /X /C "C:\work\lol(1)\maven\bin\mvn.cmd -B -e -V compile"
```

This command fails, saying that C:\work\lol is not a known command

When a space is added to the path, the path gets escaped, and the command runs successfully.

It looks like more characters need to be escaped on some platforms

---

**Affects:** maven-invoker-2.2

**Issue Links:**
- [MSHARED-765](https://issues.apache.org/jira/browse/MSHARED-765) Password not escaped (on windows?)

1 votes, 4 watchers

jira-importer:
**[Sylwester Lachiewicz](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=slachiewicz)** commented

First, it requires change in shared-utils

jira-importer:
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** commented

Problem is likely in Shell.quoteOneItem and can be tested from Shell.getCommandLine which returns quoted arguments

jira-importer:
**[Matthias Bünger](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=JIRAUSER302100)** commented

This project has moved from Jira to GitHub Issues. This issue was migrated to [apache/maven-invoker#127](https://github.com/apache/maven-invoker/issues/127). Please visit the GitHub issue to view further activity, add comments, or subscribe to receive notifications.

slachiewicz:
Closing as a duplicate of #268. Both come from the same place: `Shell.quoteOneItem` returns its input unchanged and the only quoting trigger is a space, so `cmd.exe` special characters such as `&` and `(` reach the command line unquoted. The fix belongs in this repo, not in apache/maven-invoker#127 or apache/maven-jarsigner#47, which this comment cross-links so those trackers point back here.

*This comment was created with AI assistance.*

PR #40, opened 2020-05-28
[MSHARED-297] Unconditionally single quote executable and arguments

roxspring:
- Escape arguments containing `#`, with unit test
  - Closes [MSHARED-431](https://issues.apache.org/jira/browse/MSHARED-431)
- Partial port of https://github.com/sonatype/plexus-utils/commit/b38a1b3a4352303e4312b2bb601a0d7ec6e28f41
  - BourneShell unconditionally single quotes the executable and arguments
  - ~~Shell usage is only desirable when generating code for remote execution~~ omitted because shell usage is also required for running built-in commands such as `echo` as used in the tests.
  - Closes [MSHARED-297](https://issues.apache.org/jira/browse/MSHARED-297)

PR #109, opened 2022-07-26
[MSHARED-1113] Downgrade m-resource-p to 3.1.0

slawekjaranowski:
Cause MSHARED-1112

PR #434, opened 2026-09-11
CmdShell: quote items that cmd.exe would otherwise interpret

### LINKED ISSUE #268: [MSHARED-765] Password not escaped (on windows?)
**[Claus Näveke](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=nitek)** opened **[MSHARED-765](https://issues.apache.org/jira/browse/MSHARED-765?redirect=false)** and commented

When using a password which contains the & character, the jarsigner plugin fails (in my case on windows), because the & is not escaped in the command line that is executed in background.

Workaround is to rewrite the password with ^& instead.

---

**Affects:** maven-jarsigner-1.4

**Issue Links:**
- [MSHARED-851](https://issues.apache.org/jira/browse/MSHARED-851) & must trigger quoting in Windows

- [MSHARED-832](https://issues.apache.org/jira/browse/MSHARED-832) Cannot invoke maven from path with parentheses on Windows

slachiewicz:
`CmdShell` builds one `cmd.exe /X /C "<line>"` string and never quoted the items inside it, so an executable path with parentheses or an argument with `&` was parsed by `cmd.exe` as operators. This adds `CmdShell.quoteOneItem`: an item that contains whitespace or one of the characters `cmd.exe /?` lists as special (`& < > ( ) @ ^ |`) is wrapped in double quotes; an item already wrapped in double quotes is left alone; `setQuotedArgumentsEnabled(false)` and `setUnconditionalQuoting(true)` keep their meaning.

Two things differ from the earlier attempts (#40, MSHARED-851). The quoting is conditional on the documented character set rather than unconditional, and it is confined to `CmdShell`, so `BourneShell` is untouched. It also restores something master lost in ef89c5a (#369): before that commit `Shell.quoteOneItem` double-quoted items containing a space through `quoteAndEscape`, and 3.4.2 still ships that behaviour. `Shell.getQuotingTriggerChars` was the dead remainder of that mechanism and is removed here.

Two behaviour changes to be aware of:

- The `^&` workaround from #268 stops working. `^&` contains `&`, so it is now quoted, and `^` is literal inside quotes, so the program receives `^&`. Users who applied it must drop it.
- Under `-Djdk.lang.Process.allowAmbiguousCommands=false`, Java rejects a quoted argument that contains further quotes. Previously only caller-supplied inner quotes could trigger that; now any space or special character does. Default JVMs are unaffected.

`CmdShellTest` pins the string shape on every platform. Two tests in `CommandLineUtilsTest` run only on Windows and execute a real `cmd.exe`: `echo a&b` must print `"a&b"` rather than run `b`, and a script under a `lol(1)` directory must run. The three windows-latest CI jobs are the evidence for those.

Fixes #268, which also covers the parentheses case from #286.

Verified: `mvn -B verify` on JDK 17 (macOS) -> Tests run: 800, Failures: 0, Errors: 0, Skipped: 19; spotless clean; the Windows execution tests pass on all three windows-latest jobs (JDK 8, 21, 25).

### DIFF
--- a/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
+++ b/src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
@@ -18,7 +18,9 @@
  */
 package org.apache.maven.shared.utils.cli;
 
+import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Locale;
@@ -27,6 +29,7 @@ import java.util.Properties;
 
 import org.apache.maven.shared.utils.Os;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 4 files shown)


slachiewicz:
For reviewers: how plexus-utils handles the same problem, since its `CmdShell` is the sibling of this one.

**Behaviour.** plexus-utils [`Shell.quoteOneItem`](https://github.com/codehaus-plexus/plexus-utils/blob/master/src/main/java/org/codehaus/plexus/util/cli/shell/Shell.java) goes through `quoteAndEscape` with a single trigger, the space: an item with a space gets double quotes, an already-quoted item is left alone, nothing is escaped by default. That is what shared-utils 3.4.2 did before ef89c5a removed the path. This PR restores that and adds the `cmd.exe /?` specials `& < > ( ) @ ^ |` as triggers; plexus-utils has no equivalent, so `lol(1)` and `a&b` still fail there.

**Reported there, never fixed by quoting.** [#37](https://github.com/codehaus-plexus/plexus-utils/issues/37) (asterisk in a password) has been open since 2018; [#36](https://github.com/codehaus-plexus/plexus-utils/issues/36) (embedded double quote) was closed as its duplicate. The maintainers' direction is architectural: [PR #109](https://github.com/codehaus-plexus/plexus-utils/pull/109) drops the `cmd.exe` wrapper and hands the argument vector to `ProcessBuilder`, and #37 says explicitly that the shell implementations should be replaced rather than taught more escaping. That PR has been open since 2021.

**What this PR shares with plexus-utils, on purpose.** An embedded `"` is not escaped and `*` is not quoted, because `cmd.exe` toggles quote state on every `"` and no escaping survives it reliably. Those cases belong with the "drop the shell" change, not with trigger characters. One small divergence: under unconditional quoting plexus-utils re-wraps an already-quoted item, this PR leaves it alone.

**Follow-up, separate from this PR.** plexus-utils runs `cmd.exe /X /D /C` since [9093271](https://github.com/codehaus-plexus/plexus-utils/commit/9093271) (2024, for MJAVADOC-811): `/D` skips the Command Processor AutoRun registry entry, which in some corporate setups changes the working directory of every spawned `cmd.exe`. Our `CmdShell` still runs `/X /C` and has the same exposure.

src/test/java/org/apache/maven/shared/utils/cli/CommandLineUtilsTest.java
```diff
@@ -143,6 +147,51 @@ public void executeCommandLineWithLargeStdoutCompletesWithoutFailure() throws Ex
...
+            return;
+        }
+
+        // MSHARED-832: C:\work\lol(1)\maven\bin\mvn.cmd
+        File dir = new File(tempDir, "lol(1)");
+        assertTrue(dir.mkdirs());
+        File script = new File(dir, "x.cmd");
+        Files.write(script.toPath(), "@echo ok".getBytes(StandardCharsets.US_ASCII));
+
+        Commandline cl = new Commandline();
```
elharo, line 184: cl --> commandLine

elharo (approved):

